PR #5, opened 2015-04-22
qualify node SG with instance id for deploys into master's vpc

rpattcorner:
In past nodes get SG named for their master, which works because nodes usually go in their own VPC.  Now we have nodes deploying sometimes in the master's own VPC ... sometimes because of demos, sometimes because of docker, and we encounter 

Apr 22 16:54:53 - deploy - #<Aws::EC2::Errors::InvalidGroupDuplicate: The securi                                                                                                                     ty group 'Mu Client Rules for 54.152.160.41' already exists for VPC 'vpc-bf89ccd                                                                                                                     a'>

Rather than test and retry we just qualify the Node SG by its instance ID

### DIFF
--- a/modules/mu/mommacat.rb
+++ b/modules/mu/mommacat.rb
@@ -1190,7 +1190,7 @@ MESSAGE_END
 			my_ports = [10514]
 
 			my_instance_id = MU.getAWSMetaData("instance-id")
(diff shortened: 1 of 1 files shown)


rpattcorner:
Overtaken  by events

PR #14, opened 2015-06-02
modify chef-client wrapper to warn against non-root usege

zr2d2:


jstange:
This is silly. We shouldn't be adding an extra utility, especially one that can easily break functionality, to work around trivial user error.

PR #26, opened 2015-08-26
Iw skipnagiosswitch

rpattcorner:
-s to skip nagios (with warning in help) so user actions can be invoked from within a chef run.

One odd behaviour ... after create the new user doesn't appear in the auto display of users -- but an immediately subsequent mu-user-manage -l does show the user


rpattcorner:
Tests out OK, waited a decent interval, merging

PR #32, opened 2015-10-11
AWS additions

amirahav:
## CloudWatch/CloudWatch Logs

We can now create CloudWatch alarms. Alarms can be created within each resource that supports it (eg. Databases, Servers, Server Pools, Load Balancers) or also as first class citizens in the BoK language itself. This should give you a better picture of the status of various resources. For example you can now get notified when the number of healthy instances attached to a load balancer is below a certain number, or configure AWS to "recover" you instance when it fails the System Status Check

CloudTrail monitoring was also added to CloudWatch directly. Assuming CloudTrial is enabled in your account you can enable CloudWatch integration by setting *enable_cloudtrail_logging: true* in the BoK. See example at *demo/cloudwatch-logs-cloudtrail.yaml*

## ElasticCache

You can now create cache clusters and can choose between memcached and redis. See example at *demo/cache_cluster.yaml*

The ElasticCache API is pretty awful. To make sure cleanup doesn't hold up other deployments/cleanups we are loading an existing deployment instead of searching for resources directly. This can create a situation where not all resources are deleted.

## RDS

You can now create database clusters using the Aurora engine. We will manufacture database instances for you at the configuration level based on the number of nodes specified in the *cluster_node_count* attribute (this resembles how Read Replicas are generated) . See example at *demo/db_cluster.yaml*

Databases can now be created based on a point in time state of an existing database instance. Support for point in time creation style varies between database engine types 

Password management now works the same way it does on other resources that use passwords - you can choose between a randomly generated password or a password that is stored in a vault. The option to store a clear text password in the node structure was also removed. You can retrieve the password by using a vault only. These means that both *password* and *unecrypted_master_password* that where deprecated in Internal Wrangler have been removed.

Other changes:
* Tags are now propagated to snapshots. This is not exposes at the BoK language level
* Cleanup - Will now search for other RDS resources without first looking for a database instance
* flattening of some BoK configuration attributes the were added in Internal Wrangler
* EC2-Classic support was mostly removed. Cleanup should still work, but deployment is not supported anymore
* Deployment into the default VPC should work better now, however the database instance will be set to be publicly accessible regardless of what *publicly_accessible* is set to.

This should be tested to ensure I didn't introduce any regressions.

## FlowLogs

You can now enable traffic logging on a VPC and/or subnets by setting *enable_traffic_logging: true* on the VPC or subnets. If you enable it at the VPC level it will apply to all subnets and network interfaces. If you only want to enable logging on specific subnets, set this at the subnet level.

## AutoScaling

With the addition of ClouWatch you can now create fully functioning scaling policies. You can configure multiple ClouWatch alarms to point at the same scaling policy to help you scale based on different criteria (eg. CPU usage and/or network traffic)

Instead of using the simple scaling policy type which adds/removes a fixed number of instances, you can now use step scaling which will add/remove capacity based on the magnitude of the breach. See example at *demo/autoscale_step_scaling.yaml*

# SNS

This is a very rudimentary implementation of the API, and only includes the ability to create notification topics to make is possible for CloudWatch to send notification by email.


jstange:
Sweet beans. Merging into my current working branch for testing.

amirahav:
Kewl! Thanks! Let me know what breaks.

rpattcorner:
Documentation!  Bless your heart.

We should all do this (and mostly do)

R.

Robert Patt-Corner
3865 Wilson Blvd, Suite 500.  Arlington, VA 22203.
t: (703) 652-0991  m: (410) 775-8096  f: (703) 842-7591
robert.patt-corner@eglobaltech.com<mailto:robert.patt-corner@eglobaltech.com>

[cid:image001.png@01CF7FF9.07C3F220]
www.eglobaltech.com<http://www.eglobaltech.com/>

jstange:
Minor niblet in documentation. Noting here so I don't forget to look into it later.

    [warn]: In file `modules/mu/clouds/aws/notification.rb':20: Cannot resolve link to MU::Config::BasketofKittens::notifications from text:
    [warn]: {MU::Config::BasketofKittens::notifications}...
    [warn]: In file `modules/mu/clouds/aws/notification.rb':29: Cannot resolve link to  MU::Config::BasketofKittens::notifications from text:
    [warn]: ...{MU::Config::BasketofKittens::notifications}...

mu-test-drive:
Oh heyyy, those are patch notes. Will steal.



Sent from my Verizon Wireless 4G LTE smartphone


-------- Original message --------

amirahav:
Whopsie, Sorry about that!

PR #34, opened 2015-11-17
add selinux policy for oddjobd

rpattcorner:
Add policy module for oddjobd's intereaction with syslogd, guard and execute


jstange:
No harm done to my existing instances either. Merging.

PR #40, opened 2016-03-30
Switch to SSSD for Active Directory integration on Linux

jstange:
SSSD is generally considered the correct tool nowadays, and seems overall a lot less sketchy than the cobbled-together Winbind solution.

While I was messing with Samba-related packages, I removed the dependency on GlusterFS-specific Samba packages, which have caused problems before (and none of our recipes use the Gluster CIFS support anyway).

These commits also include a few odds and ends. Gem version bumps, minor bugfixes.


jstange:
Far as I know FEMA Geoplatform is the only place in the world using our AD Linux support. I'll settle for simple smoke testing, unless someone else is also using it and can give us a more robust run-through than I already have (@amirahav if anyone).

amirahav:
Let me try it out. I do have several Linux nodes that talk to a Windows AD,
and I'm using samba-vfs-glusterfs to deal with a couple of Windows nodes that need access to Gluster

jstange:
K, be careful if any of it's important. It'll aggressively yank packages and remake things in my image.

PR #43, opened 2016-04-12
knife-windows 1.4.0

amirahav:
knife-windows 1.1.4 ended up being problematic - files in the installed gem have a Windows line ending which cause the patch to fail.
This patches knife-windows 1.4.0 instead, which uses Unix line endings.
(yes really...)

### DIFF
--- a/install/knife-windows-cygwin-1.4.0.patch
+++ b/install/knife-windows-cygwin-1.4.0.patch
@@ -1,0 +1,147 @@
+--- knife-windows-1.4.0/lib/chef/knife/bootstrap/windows-chef-client-msi.erb	2016-04-12 01:00:36.993000454 -0400
++++ knife-windows-1.4.0/lib/chef/knife/bootstrap/windows-chef-client-msi_patched.erb	2016-04-12 01:17:43.000000000 -0400
+@@ -167,6 +167,17 @@
+ 
+   <%= install_chef %>
+ 
++   SET LookForFile="c:\opscode\chef\bin\chef-client.bat"
++   @echo off
++
++   :CheckForFile
++   IF EXIST %LookForFile% GOTO FoundIt
++   c:\Windows\System32\timeout.exe /t 30
++   GOTO CheckForFile
++
++   :FoundIt
++   @echo on
++
+   @if ERRORLEVEL 1 (
+       echo Chef-client package failed to install with status code !ERRORLEVEL!. > "&2"
+       echo See installation log for additional detail: %CHEF_CLIENT_MSI_LOG_PATH%. > "&2"
+--- knife-windows-1.4.0/lib/chef/knife/bootstrap_windows_base.rb	2016-04-12 10:33:50.000000000 -0400
++++ knife-windows-1.4.0/lib/chef/knife/bootstrap_windows_base_patched.rb	2016-04-12 10:20:11.000000000 -0400
+@@ -335,7 +335,11 @@
+         # we have to run the remote commands in 2047 char chunks
+         create_bootstrap_bat_command do |command_chunk|
+           begin
+-            render_command_result = run_command(command_chunk)
++            render_command = command_chunk
++            if locate_config_value(:cygwin)
++              render_command = %q!cd $TEMP && !+command_chunk
++            end
++            render_command_result = run_command(render_command)
+             ui.error("Batch render command returned #{render_command_result}") if render_command_result != 0
+             render_command_result
+           rescue SystemExit => e
+@@ -357,11 +361,20 @@
+       end
+ 
+       def bootstrap_command
++        if locate_config_value(:cygwin)
++          @bootstrap_command ||= "cd $TEMP && cmd.exe /C #{bootstrap_bat_file}"
++        else
+         @bootstrap_command ||= "cmd.exe /C #{bootstrap_bat_file}"
+       end
++        @bootstrap_command
++      end
+ 
+       def bootstrap_render_banner_command(chunk_num)
+-        "cmd.exe /C echo Rendering #{bootstrap_bat_file} chunk #{chunk_num}"
++        if locate_config_value(:cygwin)
++          return "echo 'Rendering #{bootstrap_bat_file} chunk #{chunk_num}'"
++        else
++          return "cmd.exe /C echo Rendering #{bootstrap_bat_file} chunk #{chunk_num}"
++        end
+       end
+ 
+       def escape_windows_batch_characters(line)
+@@ -374,11 +387,18 @@
+         bootstrap_bat = ""
+         banner = bootstrap_render_banner_command(chunk_num += 1)
+         render_template(load_template(config[:bootstrap_template])).each_line do |line|
+-          escape_windows_batch_characters(line)
+           # We are guaranteed to have a prefix "banner" command that echo's chunk number.  We can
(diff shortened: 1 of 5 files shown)


jstange:
Oh, good, this solves a nuisance I've had on my list. Testing.

image

jstange:
Current Windows base images were unbuildable anyway due to patch drift, so I (successfully) built a new batch with this branch and committed those AMI ids to master. Attempting to build a simple demo off of those, though, I get the following:

    10.100.0.54 Download via PowerShell succeeded.
    10.100.0.54 Installing downloaded client package...
    10.100.0.54
    10.100.0.54 C:\Windows\TEMP>msiexec /qn /log "C:\Windows\TEMP\chef-client-msi21049.log" /i "C:\Windows\TEMP\chef-client-latest.msi"
    10.100.0.54 Successfully installed Chef Client package.
    10.100.0.54
    10.100.0.54 C:\Windows\TEMP>SET LookForFile="c:\opscode\chef\bin\chef-client.bat"
    10.100.0.54 ERROR: Input redirection is not supported, exiting the process immediately.
    10.100.0.54 ERROR: Input redirection is not supported, exiting the process immediately.
    10.100.0.54 ERROR: Input redirection is not supported, exiting the process immediately.
    10.100.0.54 ERROR: Input redirection is not supported, exiting the process immediately.

Smells like knife-windows stupidity to me. If I Ctrl-C it interrupts and restarts the bootstrap process (neat), but promptly gets in the same rut. Do you get this same behavior with **ami-018fab6b**?

amirahav:
I used ami-3d787d57 (AWS) yesterday, and a couple of custom private ones today, and haven't seen that.
Let me test that one.

jstange:
K. If you get the same weirdness, try building an AMI off of this branch from your stock, then building a machine from that and see it does the same thing. Could be something to do with detritus we're leaving around.

amirahav:
Not seeing that error, however I am seeing these with **ami-018fab6b**

    172.31.42.203 powershell.exe -ExecutionPolicy Unrestricted -NoProfile -NonInteractive -File      C:\chef\wget.ps1 "https://www.chef.io/chef/download? p=windows&pv=2012&m=&DownloadContext=PowerShell&v=12.8.1-1" "C:\Windows\TEMP\chef-client-latest.msi"
    172.31.42.203 Download via PowerShell succeeded.
    172.31.42.203 Installing downloaded client package...
    172.31.42.203
    172.31.42.203 C:\Windows\TEMP>msiexec /qn /log "C:\Windows\TEMP\chef-client-msi6793.log" /i "C:\Windows\TEMP\chef-client-latest.msi"
    172.31.42.203 Another program is being installed. Please wait until that installation is complete, and  then try installing this software again.
    WARNING: Failed to install Chef Client MSI package in remote context with status code 1618.

Omitting the stack trace because it isn't relevant  

    172.31.42.203 C:\Windows\TEMP>chef-client -c c:/chef/client.rb -j c:/chef/first-boot.json
    172.31.42.203 '"C:\opscode\chef\bin\\..\embedded\bin\ruby.exe"' is not recognized as an internal or     external command,
    172.31.42.203 operable program or batch file.
    Apr 13 00:56:19 - groomers/chef - DEMO-DEV-2016041300-SA-WINDOWS: Knife Bootstrap failed #<SystemExit: exit>, retrying (1 of 25)
    {
    Stack trace
    }


    172.31.42.203 C:\Windows\TEMP>chef-client -c c:/chef/client.rb -j c:/chef/first-boot.json
    172.31.42.203 C:/opscode/chef/embedded/lib/ruby/2.0.0/rubygems/dependency.rb:296:in `to_specs':  Could not find 'chef-config' (= 12.8.1) among 0 total gem(s) (Gem::LoadError)
    172.31.42.203   from C:/opscode/chef/embedded/lib/ruby/2.0.0/rubygems/dependency.rb:307:in `to_spec'
    172.31.42.203   from C:/opscode/chef/embedded/lib/ruby/2.0.0/rubygems/core_ext/kernel_gem.rb:47:in `gem'
    172.31.42.203   from C:/opscode/chef/bin/chef-client:4:in `<main>'
    Apr 13 00:56:51 - groomers/chef - DEMO-DEV-2016041300-SA-WINDOWS: Knife Bootstrap failed # <SystemExit: exit>, retrying (2 of 25)
    {
    Stack trace
    }

It looks like it's trying to remove the old version of chef and install the new one at the same time, so the *input redirection* error might also be related to this.
It ends up bootstrapping successfully after those errors.

jstange:
Probably a race condition then. I've noticed that the Chef install is non-blocking even when it's really supposed to be, and that the knife-windows code that tries to force it to be non-blocking just... doesn't work. Lemme know if you figure it out. I'll see how reproducible this mess is on my end.

amirahav:
I did another deploy last night with the same AMI and two more just now. The second deploy from last night bootstrapped without an issue, and from the two today one had the same issue while the other was fine, so +1 on race condition. 

I guess it can be solved in userdata. Something like checking which version of chef is installed, instead of just if it's installed. That can also be added to knife-windows.

And yes, knife-windows chef install is non-blocking - if it fails to install it will just create a scheduled task to finish the install.

amirahav:
I was able to get the **Input redirection is not supported, exiting the process immediately.**

It is caused by the **c:\Windows\System32\timeout.exe /t 30** line in _knife-windows-1.4.0/lib/chef/knife/bootstrap/windows-chef-client-msi.erb_, is not fatal, and if you wait 30 seconds bootstrap will continue.

Tried **C:/bin/cygwin/bin/sleep.exe 30** in _knife-windows-1.4.0/lib/chef/knife/bootstrap/windows-chef-client-msi.erb_ to force it to wait, but it just hangs.

I have userdata that does the chef reinstall if it's a different version, but it's irrelevant in this case because knife-windows will reinstall chef again anyway.

jstange:
Not only was my workaround too smart for a batch file, it wasn't really solving the problem, which was that the Chef installer was failing silently because we didn't properly clean up previous installers. Did some weird registry magic that I *think* got it. At any rate, tests with new Mu-based AMIs off of **cloudformation_emitter** (#44) work now.

This PR is now redundant to **cloudformation_emitter** (#44). Conduct tests on that branch and lemme know what happens.

amirahav:
Well, Windows! This only touches 32 bit installs, there is the same thing
in user data the touches 64bit as well (It will remove chef if it's a
different version than what mu says).

jstange:
I've only seen it land under Wow6432Node so far. I'll add a second clause to look in the other place.

If that's in userdata, I'm not sure why we had this problem in the first place, but whatever.

amirahav:
Look at my comment from a couple of weeks ago - regardless of what we do in userdata knife-windows will end up removing and reinstalling chef. Userdata will install the 64 bit version, while knife-windows will remove that and will currently install the 32 bit version.
Despite what it says here, specifying bootstrap_architecture had no effect when I looked at it a couple of weeks ago:
https://github.com/chef/knife-windows#specifying-the-package-architecture

I added it to userdata here: https://github.com/cloudamatic/mu/pull/43/commits/333b4be615b9c958d905380c3bd4abac8dca2dd2

jstange:
I was pondering why my fix even helps. It should be a NOOP, as your userdata is already installing something sensible and setting that guard, even if Chef bootstrap subsequently screws it up.

PR #44, opened 2016-04-21
CloudFormation Emitter

jstange:
Note: this branch also includes #40 (**Switch to SSSD for Active Directory integration on Linux**) and #43 (**knife-windows 1.4.0**).

Enhanced Parameters
----------------------------

There is now a *parameters* section available in the Basket of Kittens language. You use this to generate and require ERB-style variables to reference elsewhere in the config, much as we've always done. This helps implement some common patterns, as well as make BoKs which leverage variables more readable.

http://mu.femadata.com/docs/frames.html#!MU/Config/BasketofKittens/parameters.html

Here are some examples lifted from our Wordpress demo (https://github.com/cloudamatic/mu/blob/cloudformation_emitter/demo/wordpress.yaml), wherein we see the setting of default values, user-friendly descriptions, and several flags that interact with CloudFormation targets (more on that later).

    parameters:
      - name: appname
        default: demo
      - name: instance_type
        default: t2.micro
      - name: vpc_deploy_id
        required: false
        description: The Mu deploy identifier for an existing deploy with a single VPC, into which we'll attempt to insert this deployment.
      - name: vpc_id
        cloudtype: AWS::EC2::VPC::Id
        description: A simple VPC cloud id into which we'll attempt to insert this deployment.
        default: <%= MU.myVPC %>

The above generates ERB variables named *appname*, *instance_type*, *vpc_deploy_id*, and *vpc_id*. Note that these are bare (local) variables, as opposed to global ($foo) variables. Note also that you can still reference anything else that would been in scope for the ERB parser, e.g. *$environment*, *MU.myVPC*, etc.

Unless *required* is set to `false` or a *default* is provided, a variable is assumed to be required, and *mu-deploy* will error out if those parameters are unspecified. Note the *vpc_deploy_id* in the above example **not** being required- elsewhere in the Wordpress demo, you'll see that we check for that value, and attempt to identify a target VPC from it if *vpc_deploy_id* is set. If not, we fall back to a *vpc_id*. If that's missing, we outright build a VPC for the demo. This is a pattern we should probably encode into all of our demos, rather than our current hodgepodge (the generic constructed VPC would be an include, rather than gumming up every YAML file in the tree).

A simple enhancement might be to prompt a user for values of required parameters, if we're running from an interactive terminal. This has the potential to break @zr2d2 and @rpattcorner's various scripts, so I did not include this feature for now

CloudFormation as a target
-----------------------------------

We now have a second cloud layer... sort of. I've implemented a `MU::Cloud::CloudFormation` plugin which emits a CloudFormation stack from a BoK, instead of creating AWS resources directly. It's a fairly close approximation, and covers almost the entirety of what we cover of the AWS API in `MU::Cloud::AWS`.

You can invoke it two ways:

1. Set `cloud: CloudFormation` on one or more resources in your stack.
2. *mu-deploy -c some_bok_of_aws_resources.yaml*

By default this will dump the template to a file in /tmp and tell you where. You can use the `-l` flag to *mu-deploy* to send it somewhere else, including an S3 bucket.

Note that the intent here is **not** to provide a pointless layer of indirection to deploying Mu resources, though you could probably wrangle it to work that way (but don't). This was done to satisfy a client need, where our major deliverable is a series of templates that are not dependent on our tooling. It ended up being an enormous time-saver, so we can now lean on this whenever else we get this requirement.

Since all it really does is build a big hash, it's also a good learning tool for a new developer who's thinking about writing another cloud layer. I'd recommend anyone looking to learn the ropes look at *modules/mu/cloud/cloudformation* instead of *modules/mu/cloud/aws*, which is necessarily a lot more complex and confusing.

Parameter-related magic: When deploying to CloudFormation instead of directly to AWS, the *cloudtype*, *description*, *prettyname*, and *list_of* parameters come into play. Specifying BoK parameters will automatically generate equivalent parameters in the resultant CloudFormation templates. Here's a sample from one of my Flatworld BoKs:
    
    parameters:
    - name: instancesize
      prettyname: "Instance Size"
      default: t2.small
    - name: dnsbase
      prettyname: "Base Domain"
      default: ".flatworldknowledge.com"
      description: "Domain under which all hosts will be accessed. Used be applications to configure properly for name-based virtual hosting and to communicate with other applications in the stack."
    - name: env
      prettyname: "Environment"
      default: <%= $environment.upcase %>
    - name: scholarami
      prettyname: "Scholar AMI"
      default: ami-cb93caa1
      cloudtype: AWS::EC2::Image::Id
    - name: targetvpc
      prettyname: "Target VPC"
      default: vpc-b1340fd5
      cloudtype: AWS::EC2::VPC::Id
    - name: serversubnets
      prettyname: "Server Subnets"
      list_of: subnet_id
      default: subnet-9fb006b5,subnet-831c7df5,subnet-c88a3390,subnet-2aa3dc17
      cloudtype: List<AWS::EC2::Subnet::Id>
      description: "Subnets into which Servers will be placed. Should be private."
    - name: elbsubnets
      prettyname: "LoadBalancer Subnets"
      list_of: subnet_id
      default: subnet-9db006b7,subnet-851c7df3,subnet-d68a338e,subnet-2ba3dc16
      cloudtype: List<AWS::EC2::Subnet::Id>
      description: "Subnets into which LoadBalancer will be placed. Should be public."

And the result, when you go to launch the template from the console:

image

Cost Calculation
--------------------

As a corollary to being able to output CloudFormation, deploys should now always spit out a URL for http://calculator.s3.amazonaws.com/ that's been pre-populated with the resources from the deploy. **This is only as accurate/complete as the AWS calculator service**, which it to say not 100%. At the time of this writing, the calculator hasn't ever heard of t2.nano instances or NAT gateways. It's a convenience feature for back-of-the-napkin costing.

MIscellany
-------------

- Fixed a number of latent bugs with Alarms, ElastiCache clusters, RDS replicas, and NAT Gateways
- AWS recently changed the behavior of API endpoints with respect to setting the *encrypted* flag on  EBS volumes that originate from snapshots. Worked around that.
- A number of minor cookbook and gem versions bumped
- Many small enhancements and edge-case fixes in `MU::Config` and `MU::Deploy`. The more common mistakes we can catch at the config parser, the better.

**HAMLET**: Do you see yonder cloud that's almost in shape of a camel?
**POLONIUS**: By the mass, and 'tis like a camel, indeed.
**HAMLET**: Methinks it is like a weasel.
**POLONIUS**: It is backed like a weasel.
**HAMLET**: Or like a whale?
**POLONIUS**: Very like a whale.
**HAMLET**: Then I will come to my mother by and by.


jstange:
...this one merits extensive regression testing. I had to get my fingers into all kinds of stuff.

A couple of minor nits that don't merit holding up tests:

- Not sure the cost calc always gets run at the end of "normal" deploys.
- Need a config flag to scrub Mu-isms (extra tags, naming noise) out of CloudFormation templates.

amirahav:
@jstange I'm seeing multiple warnings/errors caused MU calling the CloudFormation layer when it shouldn't.
This usually happens at the end of deploys. As an example:

Apr 27 10:53:10 - deploy - Generating cost calculation URL for all Amazon Web Services resources.
Apr 27 10:53:17 - /opt/mu/lib/modules/mu - DNSZone parameter 'all_account_vpcs' currently not supported for CloudFormation targets
        {
  "thread": 70013738611240,
  "name": "dnszone_boundlessps.com_create"
}
Apr 27 10:53:17 - deploy - MU::MuError
        {
  "thread": 70013738611240,
  "name": "dnszone_boundlessps.com_create",
  "details": [
    "/opt/mu/lib/modules/mu/clouds/cloudformation/dnszone.rb:53:in `create'",
    "/opt/mu/lib/modules/mu/cloud.rb:950:in `call'",
    "/opt/mu/lib/modules/mu/cloud.rb:950:in `block (4 levels) in <class:Cloud>'",
    "/opt/mu/lib/modules/mu/deploy.rb:487:in `call'",
    "/opt/mu/lib/modules/mu/deploy.rb:487:in `block (2 levels) in createResources'"
  ]
}
Apr 27 10:53:17 - /opt/mu/lib/modules/mu - MU::MuError
        {
  "thread": 70013738611240,
  "name": "dnszone_boundlessps.com_create"
}
/opt/mu/lib/modules/mu/clouds/cloudformation/dnszone.rb:53:in `create': MU::MuError
        from /opt/mu/lib/modules/mu/cloud.rb:950:in `call'
        from /opt/mu/lib/modules/mu/cloud.rb:950:in `block (4 levels) in <class:Cloud>'
        from /opt/mu/lib/modules/mu/deploy.rb:487:in `call'
        from /opt/mu/lib/modules/mu/deploy.rb:487:in `block (2 levels) in createResources'

jstange:
Calling the CloudFormation layer is correct behavior, that's how the cost calculation works. Annoying messages aren't correct though, sounds like it just has to handle that more gracefully. I'll get into it.

jstange:
Fixed the cost-calculation noise, I think. It was actually a pain, with exceptions not bubbling across thread boundaries like you'd expect.

I'm trying to reproduce the race condition we've seen in #43 now, since that'd be a blocker.

jstange:
Hey @amirahav, I don't quite grok the workaround. You're checking against the value *InstanceId* in the name field, when I'd expect you'd be looking at *cloud_class* for that.

Broadly speaking, I interpret this is meaning we need to require alarms to be configured in a very specific way when targeting a *server_pool* with them, such that the dimensions will always refer to a unique value. That, or we need to convert them on-the-fly somehow. Can you rattle off in english what that should be? I'll bolt it into the config parser.

amirahav:
Hey @jstange, thought I sent you an email about that, I guess it's still in drafts :-)

The problem is that each instance in an ASG needs it's own separate alarm. That was previously done here https://github.com/cloudamatic/mu/blob/master/modules/mu/clouds/aws/server.rb#L629 based on the cloud_id of each instance, and now that I'm looking at it this was the real reason why alarm creation was in the resource itself.

What's happening now is that only one alarm is created, and as part of bootstrap each instance ends up running *setAlarm* however, because we're keeping the existing dimensions of the alarm the AWS *put_metric_alarm* call is failing because of multiple InstanceId dimensions, each pointing at a different instance.

That "workaround" is there only to make sure that things don't fall apart until it's fixed. Once that's fixed we can replace https://github.com/cloudamatic/mu/blob/cloudformation_emitter/modules/mu/clouds/aws/alarm.rb#L157 with  https://github.com/cloudamatic/mu/blob/cloudformation_emitter/modules/mu/clouds/aws/alarm.rb#L155
or, make that check smarter

jstange:
I think the simple solution is to not create the generated alarm in the config parser for server_pools. `MU::Cloud::AWS::Server.postBoot` takes care of it still, I think. I've commented that bit out, try this commit and tell me if it helps.

What other blockers, if any, are on this branch in your world? We're all clear over here.

amirahav:
Yeah, *postBoot* takes care of that, but committed a bit to make them unique. Thanks!
Still seeing some issues/errors with sssd and CentOS7 on the first run. *oddjobd* thinks it's supposed to do a reload instead of a restart. Adding *supports* to the *service* resource didn't seem to help.

jstange:
I see some commits that look like the address your blockers. 'Zat the case, or should I get into it?

amirahav:
Yeah, those do solve my blockers, or at least I haven't noticed any other issues. Good to merge from here.
This
https://github.com/cloudamatic/mu/pull/44/commits/71140ce3bfbc344040be1b49fa81f8b19e87eeb2
is probably better done in config, but that can be changed later.

jstange:
...meanwhile, the *adcli* package seems to have disappeared from EPEL in the past few days.

jstange:
Yeah, they pulled it from the el6 EPEL stream, because it's been added to RHEL 6.8. This just breaks us, of course, since CentOS doesn't have it in their own main package tree (yet?).

I modified the recipe to yank *adcli* straight from its public Git repo and compile it.

amirahav:
I wounder if it's just for 6. How about pulling that RPM and hosting it on S3 for the meantime? Will be much cleaner

jstange:
Tried that for a while, couldn't get my hands on an RPM of a proper release. A compile won't get into nutty dependency deadlocks, anyway.

I don't have any C7 deploys for it laying around, me know what you find. I suspect it's in the main package tree on that release.

amirahav:
Yeah, adcli is still in Centos7. Can you make sure it doesn't compile on 7? 
The RPM RHEL is hosting for 6 is broken? I wounder if there is a spec file around that can be used to build an RPM for 6.

jstange:
Yeah, they pulled the RPMs and SRPMs a few days ago, and all the mirrors have already picked that up. The only one I found was from an old release on a defunct mirror somewhere.

Ugh, right, this recipe is organized in annoying ways. Shuffling parts around to make sure it does the right thing on the right versions. I'll try to build FEMA's Subversion server on C7 to make sure it doesn't explode.

jstange:
Yeah, we're good. Merging the hell out of it.

PR #50, opened 2016-07-04
EFS

amirahav:
This branch has been around for ~3 months now...
## EFS

EFS volumes can now be crated in the us-west-2, us-east-1 and eu-west-1 regions, see _demo/storage_pool.yaml_ for an example.
There is also a recipe (_mu-tools::efs_) that integrates directly with the resource if you choose to use it.
It will mount the directory specified in the _directory_ attribute that is part of the _mount_points_ stanza. It defaults to /efs.
Most of the work I've done was against CentOS7, however this was tested on CentOS 6 and Amazon Linux to verify EFS can be mounted. 
## Existing deploy metadata

I'm sure @jstange isn't going to like this, and there are probably issues I haven't seen yet, but you can now add deploy metadata of an existing deploy to a new deploy. 
The reason behind this is to try and break larger stacks into smaller ones while still being able to to find resources easily on the node. This is especially useful with databases, EFS volumes and cache clusters.
Simply add the following to your servers / server pools:

```
existing_deploys:
- cloud_id: my-efs-cloud-id
  cloud_type: storage_pool
```

or

```
existing_deploys:
- mu_name: my-fully-qualified-database-mu-name
  deploy_id: my-deploy-id
  cloud_type: database`
```
## Berkshelf

Although this branch includes the  _minimal_berksfile_ branch I ended up taking another pass on this and removed all third part cookbooks. 
_Berksfile.lock_ only includes cookbooks that the internal MU cookbooks depend on explicitly.
If you aren't doing it already you will have to create a project specific _Berksfile_.
## Various fixes/changes/notes

The following covers the majority of the changes:
- There is now support for Amazon Linux (2016), simply specify `platform: amazon`. AMIs are the default Amazon ones.
- When upgrading to the latest chef-client you may see issues when chef attempts to load some gems -  to get around this you can manually remove chef client and delete the chef client embeded directory
- Removed mu-tools::maldet and mu-tools::rsyslog from the mu-node chef role. My 2 cents is that these  should be added to a project specific role instead
- When creating security groups you can now specify both SGS and IP ranges in the same rule set, eg this now works:

```
firewall_rules:
- name: my-sg
  vpc:
    deploy_id: <%= $deploy_id %>
  rules:
  - port: 2200
    hosts:
    - 0.0.0.0/0
  - port: 3000
    sgs:
    - my-sg
  - port: 80
    sgs:
    - my-sg
```
- Changes to Linux userdata to treat Amazon Linux like RHEL 6
- GlusterFS server packages are now set via attributes. Haven't tested if this actually works...
- Changed Apache ports to remove deprecated attributes. You will most likely have to manually remove existing node attributes (chef 'normal' attributes) that where saved on older MU Masters
- Fixed the new nrpe and set_local_fw recipes to not fail when MU Master doesn't have a public IP, and added an selinux policy to get nrpe to not fail on CentOS7
- set_mu_hostname now uses chef search to find nodes and only uses private IPs
- Deleted the Mu_Volume_Management IAM role. It wasn't being used and allowed every node to create volumes and attach them to any other instance.
- Tried to make several recipes more idempotent (mu-master, update_nagios_only, several others probably)
## Cleanup

This is another thing that will probably be an issue - I went through and deleted recipes/cookbooks that aren't or shouldn't be used. If you guys see something you did use - restore it.
The other intent was to merge mu-utility and mu-tools, although not much is left in mu-utility, it is still there...


rpattcorner:
Thanks, Ami!  I'd like to see us merge this, or part of this so it doesn't become stale.  Ami, John ... could you suggest a set of tests that would allow us to decide?

jstange:
@zr2d2, can you get into here and figure out where this overlaps/conflicts with your minimalist Berksfile branch in #49? If possible, it's probably sanest to merge that one into this one. As I mentioned internally last week, we need to audit what broke with the big cookbook overhaul, may as well analyze this in the process.

Testing will have to be extensive, and mostly consist of running Chef on existing deploys to see what blows up. The latest Apache cookbook confused the hell out of FEMA, for example, which I seem from your fix notes you ran into as well. The EFS stuff (awesome!) concerns me less since it's pretty standalone.

The **rsyslog** recipe belongs in the base role. Out-of-the-box log centralization is a core feature, and that's how we implement it. You can make a similar argument for **maldet**, but that's more about fulfilling a paperwork promise than anything of immediate practical value, so I care less.

The *Mu_Volume_Management* role was indeed overly broad, but it's being used by the recipes that create volumes on the fly. It'll need to be replaced with something.

I don't have a problem with deploy metadata-thieving functionality. That has obvious utility.

amirahav:
Thank you guys, much appreciated!

@jstange @zr2d2 https://github.com/cloudamatic/mu/pull/49 is already merged in here, however this goes a step further and removes all explicitly referenced third party cookbooks and only includes the custom mu ones.

Yeah, the Apache mess took a while to figure out, until I noticed there where normal attributes on the node itself, another reason to not  use normal attributes :-)

I can add the rsyslog and and maldet recipes back, however it's annoying to have to remove them manually every time. Although we do log centralization it is not based on rsyslog. 
I guess the way I think about it the base role should do as little as possible as part of a base run, and then a project can decide what they want to add on top of that. If there was an easy way to override that during bootstrap without changing the role manually that would be a good solution, however currently overriding that with attributes is problematic. 

I thought the only recipe that did on the fly volume management was the security recipe one. However it is not a base role and I don't think there should be an IAM role created by default just for that. This is something that can be done on a project that requiters that (eg. FEMA) .

jstange:
If @zr2d2's branch is in there already that's one thing out of the way. We'll have to contrive an opt-in/opt-out mechanism for the default cookbook set before we can merge this though, as removing them will definitely break a lot of stuff.

Does foodcritic or some equivalent tool bark about normal attributes, by any chance? We should really pick a standard for cookbook style at some point, ideally one with an existing enforcement tool.

The security one is part of the standard build, or was last I looked, so it should fall apart once that IAM role is removed. A better IAM role was needed anyway though. We'll need to write that out in a spare hour somewhere.

amirahav:
Not sure if foodcritic warns about normal attributes, but it will still be useful for other reasons (eg cookbook style).

I'm not seeing the security recipe as part of the standard role: https://github.com/cloudamatic/mu/blob/master/roles/mu-node.json so I don't think that role should exist by default...

jstange:
I was about to follow up and ask why you deleted the *EFS* branch, but you rolled it into *min* and the pull request is against that now. Correct?

amirahav:
The pull request was always against min :-)  I started with two branches - EFS and min but got annoyed at trying to keep them both in sync so abandoned EFS. Too make sure you guys don't accidentally work on an abandoned branch I deleted EFS.

jstange:
Playing with this today, because EFS might be relevant for a project. I'm merging **nested_cloudformation** into here and resolving a bunch of conflicts, to start. @amirahav can you comment on the removal of the *awscli*, *ec2-s3-api-tools*, *ruby-cookbook*, and *rvm* cookbooks from the local repository? I have no strong attachment to any of them, but I want to make sure we've already accounted for potential breakage.

jstange:
Sweet, @zr2d2 will get all the blame for those commits if I screwed anything up.

jstange:
After some testing, which is to say bolting EFS onto a project, I've added some capabilities to the EFS support. It can resolve inline VPCs now, which would've been a deal-breaker, and it can also do clever things if you give it *subnet_pref* now.

TODO list for this branch as I see it, after some experimentation:

- Figure out the parse issue with Tails and raw subnet_ids. That's on me.
- Add the ability to exclude the stock cookbook/role behaviors (maldet, rsyslog) for those who don't need them.
- Clarify whether we broke anything by removing some knucklehead cookbooks (*awscli*, *ec2-s3-api-tools*, *ruby-cookbook*, and *rvm*).
- Sort out a better node IAM role (or alternate mechanism) so that recipes which create/tag volumes for the local instance can still work. *mu-tools::apply_security* does indeed depend on the node being able to do its own volume management and tagging.
- I notice that cleanup on a deploy failure does not remove EFS resources. Haven't looked into why.
- We don't have an EFS implementation in the CloudFormation layer, but that's a nice-to-have, not a major issue.

amirahav:
@jstange Well, now you know I didn't test with inline VPCs :-) 

I couldn't find any references to the _awscli_, _ec2-s3-api-tools_, _ruby-cookbook_, and _rvm_ cookbooks in other cookbooks so it isn't going to break any existing MU cookbooks, however other projects may have dependencies on them. If there are any, my suggestion would be to use community cookbooks instead.

I don't remember seeing any issues with cleanup of EFS resources on deploy failure but will try looking at that over the weekend.

The mu-configure blowup has to do with Chef. I've noticed a couple of times that upgrading from older versions will end up breaking things - the vault gem being an example. In both cases I ended up removing chef client and deleting all the directories. Removing and reinstalling the gems had no effect.

Those mu-gen-docs warnings are on me - it's part of adding cloudwatch and calculated DNS health checks. _healthcheck_ is now an array and called _healthchecks_.
This is currently not smart enough to parse inline alarms, and I probably won't be able to get to that before the first or second week in September.

So when using _subnet_pref_ in EFS will it resolve to unique AZs across all mountpints or just unique subnets? or neither?

jstange:
@amirahav: I think you're right re the cookbooks. The project I was working on yesterday was bollocksed by the missing *awscli* cookbook, but the solution was trivial. I just added a Berksfile at the root of that repo, told it about the cookbooks in said repo, and it worked it out from there. That is, in fact, a scriptable solution we could do in *mu-configure* if we felt like it. Generating platform repo Berksfiles, that is.

Regular *mu-cleanup* seems to take care of EFS resources just fine, it was just the interrupted deploys that didn't clean it up properly. Not sure why they'd behave differently.

I had a similar experience fixing the vault gem. It may be worthwhile to make the upgrade procedure just nuke the existing Chef client installation (like, `rpm -e chef ; rm -rf /opt/chef`). This seems to help with the sluggishness we see loading Chef's gems on older masters, too.

So *subnet_pref* will first resolve the set of subnets you think it means, like it does with every other resource (default is *all_private*). Then it replicates your mount point declaration so that you have one for each matching subnet. It appends a simple counter to the end of the names to differentiate them for when there's more than one. It's basically taking the ERB pattern in your sample BoK and shifting it into the config parser.

I did something similar with VPCs recently- now if you don't declare any subnets, it'll assume you mean the standard pattern of "1 public + 1 private per AZ" and just generate that for you.

By the way, you had some orphaned code in here. It looked like you were going to set up a *storage*pool* key under *servers* and *server_pools*, then abandoned it because it didn't actually do anything of value. Do I have that right?

Hammering out a few bugs before I go do the work I'm supposed to do today, expect another commit in a bit.

jstange:
@amirahav I can't seem to reproduce your subnet_id string concatenation problem. Is it resolved with current code?

Your *storage_pool* demo has some issues. It explicitly harvests the subnets in the Mu master's VPC, but then also takes a VPC parameter. Soon as you specify something else with the latter, the former cease making any sense. I stripped out the subnet logic rather than rewrite, since the config parser does it now anyway.

jstange:
Turns out I might need it on a project, so I got to bill some time making some *mu-tools* recipes skippable. Like this:

    application_attributes:
      skip_recipes:
      - maldet
      - clamav
      - rsyslog
      - base_repositories
      - set_mu_hostname
      - nrpe
      - split_var_partitions
      - apply_security

There's a structural change that goes with this: *role[mu-node]* is no longer invoked in the `knife bootstrap` phase of things, but rather in the main Chef run along with everything else. This solved two problems- `knife bootstrap` Chef runs ignore node attributes for some reason, and the fact that knucklehead stuff like Maldet's SSL certificate being broken could blow up deploys by breaking bootstraps.

amirahav:
@jstange Yes, generating project specific Berksfiles is right way to go, we really shouldn't be adding any third party cookbooks to the mu default Berksfile, instead we just need to make sure there is a version constraint on the default mu cookbooks when needed.

I did try interrupting an EFS deploy and causing a cleanup - it did try to clean it up, but failed because the creation call threw an error when it was trying to clean it. 
This looks like a threading issue, which is probably caused by something stoopid that I did.

Just looked at how it's resolving subnets in EFS and I like it. 
Just a quick note on this - it doesn't make sense to create one mount point per private/public subnet, instead there should be one mount point per AZ.

The storage pool demo was intended to be stoopid and only work on the VPC mu was in, your fixes make it a lot more useful for other projects

Correct about the _storagepool_ key in _servers_ and _server_pools_. It's abandoned code, but I thought I cleaned that up - guess not. 

Yes, I can still reproduce the subnet_id string concatenation problem. Just tired it with your latest updates and still seeing this:
```Aug 06 11:13:15 - /opt/mu/lib/modules/mu - Failed to locate subnet from {"subnet_id"=>"subnet-9c3751a6subnet-9c3751a6"} in LoadBalancer lb
        {
  "thread": 12940360,
  "name": "loadbalancer_lb_create"
}
Aug 06 11:13:15 - deploy - MU::MuError
        {
  "thread": 12940360,
  "name": "loadbalancer_lb_create",
  "details": [
    "/opt/mu/lib/modules/mu/clouds/aws/loadbalancer.rb:91:in `block in create'",
    "/opt/mu/lib/modules/mu/clouds/aws/loadbalancer.rb:88:in `each'",
    "/opt/mu/lib/modules/mu/clouds/aws/loadbalancer.rb:88:in `create'",
    "/opt/mu/lib/modules/mu/cloud.rb:962:in `call'",
    "/opt/mu/lib/modules/mu/cloud.rb:962:in `block (4 levels) in <class:Cloud>'",
    "/opt/mu/lib/modules/mu/deploy.rb:503:in `call'",
    "/opt/mu/lib/modules/mu/deploy.rb:503:in `block (2 levels) in createResources'"
  ]
}```

Love the optional _mu-tools_ recipes and moving them to the main chef run. The base chef run should do as little as possible, mostly to lessen the chance of breakage at that stage, and because you obviously can't override node attributes at that stage without modifying the _mu-tools_  recipes or the mu-node role themselves.

**Comment by jstange** [comment]
Did a bit of research on how to file down that IAM role. To recap, existing recipes need nodes to be able to:

- create volumes
- tag those volumes
- attach those volumes to themselves

Ami yanked the generic privileges off of nodes in this branch, and for the compelling reason that those permissions allow nodes to manipulate any volumes they damn well please, whether they belong to them or not. Only that third operation, attachment, can be properly restricted by IAM to the calling instance. Which leaves us with a bit of a conundrum.

My proposed solution: A `MommaCat` API call similar to the one asynchronous grooms use to sign SSL certificates. A node, or more pointedly a recipe running on a node, would phone in and just ask Momma to build, tag, and attach a volume on its behalf.

**Comment by jstange** [comment]
In fixing a problem @zr2d2 was having, which turns out to be AWS quietly squashing requests for multiple mount points in a single AZ, I think I discovered the hard-to-find cleanup bug: orphaned ENIs. They may have only been an issue in the scenario I just dealt with, but that's a thing to look out for. We may end up having to explicitly purge them inside `MU::Cloud::AWS::StoragePool.cleanup`.

**Comment by jstange** [comment]
@amirahav that subnet_id duplication problem looks like it was just me doing `<<` to a string instead of `=`. Tell me if the latest commit doesn't fix that.

**Comment by amirahav** [comment]
@jstange Yup, looks good thanks!
Are you seeing orphaned ENIs only with storage pools?

**Comment by jstange** [comment]
Yeah, but I'm half-sure it was an artifact of the thing I fixed yesterday. You're not supposed to create multiple mount points in the same AZ, but if you try to, instead of blowing up in your face like it should, it just kind of quietly... doesn't do it. I think I was seeing the creation of ENIs for mount targets that never actually came into being, and which subsequently didn't get purged by cleanup. Fixing the config parser to yell at you if you try to do that may have effectively fixed that as well. I can't reproduce it now, at least.

Unrelated problem: I've noticed that exceptions in the middle of deploys on this branch are outright killing `mu-deploy` instead of letting it clean up first. Did you meddle with exception handling or thread variables at all? I don't get this on any other codebase.

ex:

    Received SIGINT, hit ctrl-C again within five seconds to kill this deployment.
    ^CSep 20 09:09:51 - deploy - #<RuntimeError: Terminated by user>
            {
      "thread": 7230420,
      "name": "main_thread",
      "details": [
        "/opt/mu/lib/modules/mu/deploy.rb:181:in `block in run'",
        "/opt/mu/lib/modules/mu/deploy.rb:253:in `call'",
        "/opt/mu/lib/modules/mu/deploy.rb:253:in `join'",
        "/opt/mu/lib/modules/mu/deploy.rb:253:in `block in run'",
        "/opt/mu/lib/modules/mu/deploy.rb:252:in `each'",
        "/opt/mu/lib/modules/mu/deploy.rb:252:in `run'",
        "/opt/mu/bin/mu-deploy:136:in `<main>'"
      ]
    }
    Sep 20 09:09:51 - cleanup - Checking for cloud resources from EFS-DEV-2016092009-RD in ap-south-1
    Sep 20 09:09:51 - cleanup - Checking for cloud resources from EFS-DEV-2016092009-RD in ap-southeast-2
    Sep 20 09:09:51 - cleanup - Checking for cloud resources from EFS-DEV-2016092009-RD in eu-central-1
    Sep 20 09:09:51 - cleanup - Checking for cloud resources from EFS-DEV-2016092009-RD in ap-northeast-1
    Sep 20 09:09:51 - cleanup - Checking for cloud resources from EFS-DEV-2016092009-RD in us-east-1
    Sep 20 09:09:51 - cleanup - Checking for cloud resources from EFS-DEV-2016092009-RD in sa-east-1
    Sep 20 09:09:51 - cleanup - Checking for cloud resources from EFS-DEV-2016092009-RD in us-west-1
    Sep 20 09:09:51 - cleanup - Checking for cloud resources from EFS-DEV-2016092009-RD in eu-west-1
    Sep 20 09:09:51 - cleanup - Checking for cloud resources from EFS-DEV-2016092009-RD in ap-southeast-1
    Sep 20 09:09:51 - cleanup - Checking for cloud resources from EFS-DEV-2016092009-RD in us-west-2
    Sep 20 09:09:51 - cleanup - Checking for cloud resources from EFS-DEV-2016092009-RD in ap-northeast-2
    Sep 20 09:09:51 - clouds/aws/storage_pool - Deleting mount target fsmt-1238e95b for filesystem EFS-DEV-2016092009-RD-ARF: fs-3c9e5775
    Sep 20 09:10:01 - /opt/mu/lib/modules/mu - Failed to create mount target fsmt-1238e95b
    /opt/mu/lib/modules/mu/clouds/aws/storage_pool.rb:212:in `block in create_mount_target': MU::MuError
            from /opt/mu/lib/modules/mu/clouds/aws/storage_pool.rb:205:in `loop'
            from /opt/mu/lib/modules/mu/clouds/aws/storage_pool.rb:205:in `create_mount_target'
            from /opt/mu/lib/modules/mu/clouds/aws/storage_pool.rb:86:in `block (2 levels) in create'
    [root@femadata-mu gis]#

**Comment by jstange** [comment]
Updated TODO list for this branch:

- Give the new MommaCat `add_volume` functionality a shakedown and deprecate the old `CAPVolume` Chef library. (@jstange)
- Figure out why interrupted deploys aren't finishing cleanup correctly. (@amirahav)
- Document the little support library that's growing in `cookbooks/mu-tools/libraries/helper.rb`, maybe use it to gloss over some repetitive Chef code elsewhere. (@jstange)
- Hide Scratchpad secrets behind some JavaScript so that peoples' dumb spam filters don't destroy secrets by prefetching them. (@zr2d2)
- Update or rewrite `mu-utility::nat` to use new firewall resources. This recipe is actually broken for unrelated reasons right now-  we need to explicitly add `ACCEPT` rules to the `FORWARD` chain for the network behind the NAT now, which wasn't necessary in the past. May as well modernize this recipe while we're at it.
- (OPTIONAL) Bang out a `StoragePool` implementation for the CloudFormation layer.

I think the `existing_deploy` functionality is currently a little half-baked, but that's a rabbit hole we don't need to dive into for this thing to be merge-ready.

**Comment by jstange** [comment]
Different symptom the interrupt/cleanup bug: Failed DNS updates on this branch can kill MommaCat. This code kills all the threads when an `InvalidChangeBatch` happens, even though that's obviously not what it's supposed to do:

          begin
            change_id = MU::Cloud::AWS.route53.change_resource_record_sets(params).change_info.id
          rescue Aws::Route53::Errors::PriorRequestNotComplete => e
            sleep 10
            retry
          rescue Aws::Route53::Errors::InvalidChangeBatch, Aws::Route53::Errors::InvalidInput, Exception => e
            return if e.message.match(/ but it already exists$/) and !delete
            MU.log "Failed to change DNS records, #{e.inspect}", MU::ERR, details: params
            raise e if !delete
            MU.log "Record #{name} (#{type}) in #{id} can't be deleted. Already removed? #{e.inspect}", MU::WARN, details: params
            return
          end


```
Sep 23 14:06:01 - clouds/aws/dnszone - Failed to change DNS records, #<Aws::Route53::Errors::InvalidChangeBatch: RRSet of type A with DNS name rdgw2.femadata.com. is not permitted because a conflicting RRSet of type  CNAME with the same DNS name already exists in zone femadata.com.>
        {
  "details": {
    "hosted_zone_id": "/hostedzone/ZHTSR0VUKH1FC",
    "change_batch": {
      "changes": [
        {
          "action": "UPSERT",
          "resource_record_set": {
            "name": "rdgw2.femadata.com.",
            "type": "A",
            "ttl": 7200,
            "resource_records": [
              {
                "value": "54.208.180.175"
              }
            ]
          }
        }
      ]
    }
  }
}
Exiting!
/opt/rubies/ruby-2.1.6/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.5.5/lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call': RRSet of type A with DNS name rdgw2.femadata.com. is not permitted because a conflicting RRSet of type  CNAME with the same DNS name already exists in zone femadata.com. (Aws::Route53::Errors::InvalidChangeBatch)
        from /opt/rubies/ruby-2.1.6/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.5.5/lib/aws-sdk-core/plugins/param_converter.rb:20:in `call'
        from /opt/rubies/ruby-2.1.6/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.5.5/lib/seahorse/client/plugins/response_target.rb:21:in `call'
        from /opt/rubies/ruby-2.1.6/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.5.5/lib/seahorse/client/request.rb:70:in `send_request'
        from /opt/rubies/ruby-2.1.6/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.5.5/lib/seahorse/client/base.rb:207:in `block (2 levels) in define_operation_methods'
        from /opt/mu/lib/modules/mu/clouds/aws.rb:364:in `call'
        from /opt/mu/lib/modules/mu/clouds/aws.rb:364:in `method_missing'
        from /opt/mu/lib/modules/mu/clouds/aws/dnszone.rb:512:in `manageRecord'
        from /opt/mu/lib/modules/mu/clouds/aws/dnszone.rb:233:in `block (2 levels) in createRecordsFromConfig'
        from /opt/rubies/ruby-2.1.6/lib/ruby/gems/2.1.0/gems/logging-2.1.0/lib/logging/diagnostic_context.rb:450:in `call'
        from /opt/rubies/ruby-2.1.6/lib/ruby/gems/2.1.0/gems/logging-2.1.0/lib/logging/diagnostic_context.rb:450:in `block in create_with_logging_context
```

**Comment by jstange** [comment]
...hold on, that might not be related, checking...

**Comment by jstange** [comment]
Nah, it is. Notice that the stack trace stops at a thread boundary, instead of going all the way up to one of the many parent blocks which would actually catch this exception (at worst, all the way up in `mommacat.ru`. That ain't right.

**Comment by amirahav** [comment]
Hmmm, I don't think I changed that but I'm sure I did something stoopid. Will look at it over the weekend

**Comment by jstange** [comment]
Could easily be related to weirdness in other places. Anything that refers to `Thread` is suspect, for example. Gem updates, Ruby updates, etc...

**Comment by amirahav** [comment]
@jstange So at least on the storage pool side interrupted deploy cleanup should be working now. Will look at DNS a bit later.
However I see the following when deploying storage pools. This seems to be coming from the cloudformation layer:

    mu-deploy demo/storage_pool.yaml
    Sep 27 01:16:18 - mu-deploy - Loading /opt/mu/lib/demo/storage_pool.yaml
    Sep 27 01:16:18 - config - Guessing that /opt/mu/lib/demo/storage_pool.yaml is YAML based on parser
    Sep 27 01:16:18 - config - Passing variable 'instance_type' into /opt/mu/lib/demo/storage_pool.yaml with value 't2.small'
    Sep 27 01:16:18 - config - Passing variable 'vpc' into /opt/mu/lib/demo/storage_pool.yaml with value 'vpc-8bf71aee'
    Sep 27 01:16:18 - config - Guessing that /opt/mu/lib/demo/storage_pool.yaml is YAML based on parser
    Sep 27 01:16:24 - deploy - Deployment id: EFS "Sphinx Panther" (EFS-DEV-2016092701-SP)
    Sep 27 01:16:25 - mommacat - Creating deploy secret for EFS-DEV-2016092701-SP
    Sep 27 01:16:26 - clouds/aws/firewall_rule - Creating EC2 Security Group EFS-DEV-2016092701-SP-STORAGE-MOUNTPOINT
    Sep 27 01:16:27 - clouds/aws/storage_pool - Creating storage pool EFS-DEV-2016092701-SP-DEMO
    Sep 27 01:16:27 - clouds/aws/storage_pool - Waiting for EFS-DEV-2016092701-SP-DEMO: fs-6b985022 to become available
    Sep 27 01:16:31 - clouds/aws/firewall_rule - Setting rules in Security Group EFS-DEV-2016092701-SP-STORAGE-MOUNTPOINT (sg-2f743b55)
    Sep 27 01:16:38 - clouds/aws/storage_pool - Creating mount target for filesystem fs-6b985022
    Sep 27 01:16:38 - clouds/aws/storage_pool - Waiting for fsmt-983fefd1 to become available
    Sep 27 01:18:18 - clouds/aws/storage_pool - Waiting for fsmt-983fefd1 to become available
    Sep 27 01:19:25 - clouds/aws - Using AWS Security Group 'Mu Client Rules for 54.173.81.59' (sg-93d75ef4)
    Sep 27 01:19:25 - clouds/aws - Revoking old rules for port 10514 from sg-93d75ef4
    Sep 27 01:19:25 - clouds/aws - Adding current IP list to allow rule for port 10514 in sg-93d75ef4
    Sep 27 01:19:25 - deploy - Generating cost calculation URL for all Amazon Web Services resources.
    Sep 27 01:19:26 - clouds/cloudformation - Review estimated monthly cost for AWS resources in this stack: http://calculator.s3.amazonaws.com/calc5.html?key=cloudformation/dc201362-23bf-4156-9ec0-55ad23df826c
    /opt/mu/lib/modules/mu/cloud.rb:287:in `rescue in loadCloudType': MU::Cloud::MuCloudResourceNotImplemented (MU::Cloud::MuCloudResourceNotImplemented)
            from /opt/mu/lib/modules/mu/cloud.rb:281:in `loadCloudType'
            from /opt/mu/lib/modules/mu/cloud.rb:416:in `initialize'
            from /opt/mu/lib/modules/mu/deploy.rb:489:in `new'
            from /opt/mu/lib/modules/mu/deploy.rb:489:in `block (2 levels) in createResources'


And as for Samaba/SSSD/Winbind - could you see if the changes I committed don't break anything on your end?

**Comment by amirahav** [comment]
Not seeing an issue on DNS interrupt/cleanup as part of a server deployment:


    Sep 27 19:33:22 - mu-deploy - Loading /root/test.yaml
    Sep 27 19:33:22 - config - Guessing that /root/test.yaml is YAML based on filename
    Sep 27 19:33:22 - config - Passing variable 'vpc' into /root/test.yaml with value 'vpc-8bf71aee'
    Sep 27 19:33:22 - config - Guessing that /root/test.yaml is YAML based on filename
    Sep 27 19:33:27 - deploy - Deployment id: TETSDNS "Siamese Tacit" (TETSDNS-DEV-2016092719-ST)
    Sep 27 19:33:27 - mommacat - Generating SSH key deploy-TETSDNS-DEV-2016092719-ST
    Sep 27 19:33:36 - mommacat - Creating deploy secret for TETSDNS-DEV-2016092719-ST
    Sep 27 19:33:37 - clouds/aws/firewall_rule - Creating EC2 Security Group TETSDNS-DEV-2016092719-ST-ADMIN-VPC-8BF71AEE
    Sep 27 19:33:37 - clouds/aws/firewall_rule - sg-a45018de not yet ready, waiting...
    Sep 27 19:33:48 - clouds/aws/firewall_rule - Setting rules in Security Group TETSDNS-DEV-2016092719-ST-ADMIN-VPC-8BF71AEE (sg-a45018de)
    Sep 27 19:33:51 - clouds/aws/server - Creating IAM role and policies for 'MU::Cloud::AWS::Server' nodes
    Sep 27 19:33:51 - clouds/aws/server - Merging policy Mu_Bootstrap_Secret_TETSDNS-DEV-2016092719-ST into TETSDNS-DEV-2016092719-ST-TEST
            {
      "details": "{\"Version\":\"2012-10-17\",\"Statement\":[{\"Effect\":\"Allow\",\"Action\":[\"s3:GetObject\"],\"Resource\":\"arn:aws:s3:::ps-chef-server/TETSDNS-DEV-2016092719-ST-secret\"}]}"
    }
    Sep 27 19:33:51 - clouds/aws/server - Deploying TETSDNS-DEV-2016092719-ST-TEST into VPC vpc-8bf71aee Subnet subnet-9c3751a6
    Sep 27 19:33:52 - clouds/aws/server - Creating EC2 instance TETSDNS-DEV-2016092719-ST-TEST
    Sep 27 19:34:02 - clouds/aws/server - TETSDNS-DEV-2016092719-ST-TEST (i-ab7f3553) coming online
    Sep 27 19:34:08 - clouds/aws/server - Waiting for EC2 instance TETSDNS-DEV-2016092719-ST-TEST to be ready...
    Sep 27 19:34:52 - clouds/aws/dnszone - Adding DNS record tetsdns-dev-2016092719-st-test.server.i-2d63d5ff.platform-mu. => ["172.31.59.113"] (A) to /hostedzone/Z6R19TME74M3V
    Sep 27 19:34:52 - clouds/aws/dnszone - Adding DNS record geoserver-cluster.boundlessps.com. => ["ec2-54-85-170-102.compute-1.amazonaws.com"] (CNAME) to /hostedzone/Z1GSD5Y2STG3W4
    Sep 27 19:34:52 - clouds/aws/dnszone - Failed to change DNS records, #<Aws::Route53::Errors::InvalidChangeBatch: RRSet of type CNAME with DNS name geoserver-cluster.boundlessps.com. is not permitted as it conflicts with other records with the same DNS name in zone boundlessps.com.>
            {
      "details": {
        "hosted_zone_id": "/hostedzone/Z1GSD5Y2STG3W4",
        "change_batch": {
          "changes": [
            {
              "action": "UPSERT",
              "resource_record_set": {
                "name": "geoserver-cluster.boundlessps.com.",
                "type": "CNAME",
                "ttl": 30,
                "resource_records": [
                  {
                    "value": "ec2-54-85-170-102.compute-1.amazonaws.com"
                  }
                ]
              }
            }
          ]
        }
      }
    }
    Sep 27 19:34:52 - deploy - #<Aws::Route53::Errors::InvalidChangeBatch: RRSet of type CNAME with DNS name geoserver-cluster.boundlessps.com. is not permitted as it conflicts with other records with the same DNS name in zone boundlessps.com.>
            {
      "thread": 12243920,
      "name": "main_thread",
      "details": [
        "/opt/rubies/ruby-2.1.6/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.5.5/lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call'",
        "/opt/rubies/ruby-2.1.6/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.5.5/lib/aws-sdk-core/plugins/param_converter.rb:20:in `call'",
        "/opt/rubies/ruby-2.1.6/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.5.5/lib/seahorse/client/plugins/response_target.rb:21:in `call'",
        "/opt/rubies/ruby-2.1.6/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.5.5/lib/seahorse/client/request.rb:70:in `send_request'",
        "/opt/rubies/ruby-2.1.6/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.5.5/lib/seahorse/client/base.rb:207:in `block (2 levels) in define_operation_methods'",
        "/opt/mu/lib/modules/mu/clouds/aws.rb:364:in `call'",
        "/opt/mu/lib/modules/mu/clouds/aws.rb:364:in `method_missing'",
        "/opt/mu/lib/modules/mu/clouds/aws/dnszone.rb:512:in `manageRecord'",
        "/opt/mu/lib/modules/mu/clouds/aws/dnszone.rb:233:in `block (2 levels) in createRecordsFromConfig'"
      ]
    }
    Sep 27 19:34:52 - cleanup - Checking for cloud resources from TETSDNS-DEV-2016092719-ST in ap-southeast-2
    Sep 27 19:34:52 - cleanup - Checking for cloud resources from TETSDNS-DEV-2016092719-ST in ap-northeast-2
    Sep 27 19:34:52 - cleanup - Checking for cloud resources from TETSDNS-DEV-2016092719-ST in ap-northeast-1
    Sep 27 19:34:52 - cleanup - Checking for cloud resources from TETSDNS-DEV-2016092719-ST in sa-east-1
    Sep 27 19:34:52 - cleanup - Checking for cloud resources from TETSDNS-DEV-2016092719-ST in eu-central-1
    Sep 27 19:34:52 - cleanup - Checking for cloud resources from TETSDNS-DEV-2016092719-ST in us-east-1
    Sep 27 19:34:52 - cleanup - Checking for cloud resources from TETSDNS-DEV-2016092719-ST in us-west-2
    Sep 27 19:34:52 - cleanup - Checking for cloud resources from TETSDNS-DEV-2016092719-ST in ap-south-1
    Sep 27 19:34:52 - cleanup - Checking for cloud resources from TETSDNS-DEV-2016092719-ST in us-west-1
    Sep 27 19:34:52 - cleanup - Checking for cloud resources from TETSDNS-DEV-2016092719-ST in ap-southeast-1
    Sep 27 19:34:52 - cleanup - Checking for cloud resources from TETSDNS-DEV-2016092719-ST in eu-west-1
    Sep 27 19:34:53 - clouds/aws/server - Removing IAM role and policies for 'TETSDNS-DEV-2016092719-ST-TEST'
    Sep 27 19:34:53 - clouds/aws/server - Cleaning up ["54.85.170.102", "172.31.59.113"] from /root/.ssh/known_hosts
    Sep 27 19:34:53 - clouds/aws/server - Cleaning up ["54.85.170.102", "172.31.59.113"] from /opt/mu/var/nagios_user_home/.ssh/known_hosts
    Sep 27 19:34:53 - clouds/aws/server - Terminating i-ab7f3553 (TETSDNS-DEV-2016092719-ST-TEST) false
    Sep 27 19:34:56 - cleanup - Deleting key pair deploy-TETSDNS-DEV-2016092719-ST from eu-west-1
    Sep 27 19:34:56 - cleanup - Deleting key pair deploy-TETSDNS-DEV-2016092719-ST from us-west-1
    Sep 27 19:34:56 - cleanup - Deleting key pair deploy-TETSDNS-DEV-2016092719-ST from eu-central-1
    Sep 27 19:34:56 - cleanup - Deleting key pair deploy-TETSDNS-DEV-2016092719-ST from ap-northeast-2
    Sep 27 19:34:56 - cleanup - Deleting key pair deploy-TETSDNS-DEV-2016092719-ST from ap-southeast-1
    Sep 27 19:34:56 - cleanup - Deleting key pair deploy-TETSDNS-DEV-2016092719-ST from us-west-2
    Sep 27 19:34:57 - cleanup - Deleting key pair deploy-TETSDNS-DEV-2016092719-ST from ap-south-1
    Sep 27 19:34:57 - cleanup - Deleting key pair deploy-TETSDNS-DEV-2016092719-ST from ap-southeast-2
    Sep 27 19:34:57 - cleanup - Deleting key pair deploy-TETSDNS-DEV-2016092719-ST from ap-northeast-1
    Sep 27 19:34:57 - cleanup - Deleting key pair deploy-TETSDNS-DEV-2016092719-ST from sa-east-1
    Sep 27 19:37:54 - clouds/aws/server - i-ab7f3553 (TETSDNS-DEV-2016092719-ST-TEST) terminated
    Sep 27 19:37:54 - clouds/aws/firewall_rule - Revoking rules in EC2 Security Group TETSDNS-DEV-2016092719-ST-ADMIN-VPC-8BF71AEE (sg-a45018de)
    Sep 27 19:37:55 - clouds/aws/firewall_rule - Removing EC2 Security Group TETSDNS-DEV-2016092719-ST-ADMIN-VPC-8BF71AEE
    Sep 27 19:37:55 - cleanup - Deleting key pair deploy-TETSDNS-DEV-2016092719-ST from us-east-1
    Sep 27 19:37:55 - groomers/chef - Loading Chef libraries as user mu...
    Sep 27 19:38:00 - groomers/chef - Chef libraries loaded (took 5.353235898 seconds)
    Sep 27 19:38:00 - mommacat - Purging /opt/mu/var/deployments/TETSDNS-DEV-2016092719-ST
    Sep 27 19:38:00 - cleanup - Moving /root/.ssh/deploy-TETSDNS-DEV-2016092719-ST to /root/.ssh/archive/deploy-TETSDNS-DEV-2016092719-ST
    Sep 27 19:38:01 - clouds/aws - Using AWS Security Group 'Mu Client Rules for 54.173.81.59' (sg-93d75ef4)
    Sep 27 19:38:01 - clouds/aws - Revoking old rules for port 10514 from sg-93d75ef4
    Sep 27 19:38:01 - clouds/aws - Adding current IP list to allow rule for port 10514 in sg-93d75ef4
    Sep 27 19:38:01 - mommacat - Updating /opt/mu/var/nagios_user_home/.ssh/config...
    Sep 27 19:38:13 - mommacat - Updating Nagios monitoring config, this may take a while...
    Sep 27 19:38:39 - mommacat - Nagios monitoring config update complete.

**Comment by jstange** [comment]
That error you get in the Cloudformation layer is correct- what's bogus is that's it's supposed to catch it and display a meaningful message instead of dropping dead. There's a rescue somewhere in deploy.rb that does it, but we're dying somewhere deeper instead of bubbling up.

Ditto the DNS thing. As it occurs in Momma Cat grooms in FEMA world, that should be non-fatal, but we're killing all the threads on this branch.

**Comment by amirahav** [comment]
I'm still not seeing anything wrong with how threads are getting killed...

https://github.com/cloudamatic/mu/pull/50/commits/3eb67c4d0056875f4e894611597ca74a3f6a0a0b fixes the cost estimate/CloudFormation issue. 
As for DNS - I'm seeing the expected fatal error on initial deploy (see comment from earlier today. The deploy gets cleaned up correctly) and the expected non-fatal on re-grooming:

    Sep 27 23:14:02 - mommacat - Processing request from 54.84.248.227 (MU-ID TETSDNS-DEV-2016092722-CR, server: test, instance: i-9ccf208a, mu_ssl_sign: , mu_user root, path /)
    Sep 27 23:14:02 - mommacat - Matched ciphertext for TETSDNS-DEV-2016092722-CR
    Sep 27 23:14:02 - mommacat - Found an existing node named TETSDNS-DEV-2016092722-CR-TEST
    Sep 27 23:14:02 - mommacat - Re-grooming TETSDNS-DEV-2016092722-CR-TEST
    Sep 27 23:14:03 - clouds/aws/dnszone - Adding DNS record geoserver-cluster.boundlessps.com. => ["ec2-54-84-248-227.compute-1.amazonaws.com"] (CNAME) to /hostedzone/Z1GSD5Y2STG3W4
    Sep 27 23:14:03 - clouds/aws/dnszone - Failed to change DNS records, #<Aws::Route53::Errors::InvalidChangeBatch: RRSet of type CNAME with DNS name geoserver-cluster.boundlessps.com. is not permitted as it conflicts with other records with the same DNS name in zone boundlessps.com.>
            {
      "details": {
        "hosted_zone_id": "/hostedzone/Z1GSD5Y2STG3W4",
        "change_batch": {
          "changes": [
            {
              "action": "UPSERT",
              "resource_record_set": {
                "name": "geoserver-cluster.boundlessps.com.",
                "type": "CNAME",
                "ttl": 30,
                "resource_records": [
                  {
                    "value": "ec2-54-84-248-227.compute-1.amazonaws.com"
                  }
                ]
              }
            }
          ]
        }
      }
    }
    Sep 27 23:14:07 - cloud - Attempting SSH to 172.31.26.13 (TETSDNS-DEV-2016092722-CR-TEST) as centos with key /root/.ssh/deploy-TETSDNS-DEV-2016092722-CR
    Sep 27 23:14:07 - clouds/aws/server - Node TETSDNS-DEV-2016092722-CR-TEST has already been bootstrapped, skipping groomer setup.
    Sep 27 23:14:08 - groomers/chef - Saving TETSDNS-DEV-2016092722-CR-TEST Chef artifacts
    Sep 27 23:14:08 - groomers/chef - Granting TETSDNS-DEV-2016092722-CR-TEST access to splunk admin_user
    Sep 27 23:14:12 - groomers/chef - Saving TETSDNS-DEV-2016092722-CR-TEST Chef artifacts
    Sep 27 23:14:13 - groomers/chef - Saving TETSDNS-DEV-2016092722-CR-TEST Chef artifacts
    Sep 27 23:14:14 - groomers/chef - Saving TETSDNS-DEV-2016092722-CR-TEST Chef artifacts
    Sep 27 23:14:15 - groomers/chef - Invoking Chef on TETSDNS-DEV-2016092722-CR-TEST: Full Initial Run
    Sep 27 23:14:15 - cloud - Attempting SSH to 172.31.26.13 (TETSDNS-DEV-2016092722-CR-TEST) as centos with key /root/.ssh/deploy-TETSDNS-DEV-2016092722-CR

**Comment by amirahav** [comment]
I see it now on the third or fourth re-grooming...

**Comment by amirahav** [comment]
https://github.com/cloudamatic/mu/commit/7654bf8920d8b71847bc7e06121097fdcddc4af6 fixes the DNS issue, so I don't think we have an issue with all threads getting killed...

**Comment by jstange** [comment]
I'm guessing that 574d718 includes some stuff to tell Berkshelf to upload things itself? It's breaking the hell out of things in FEMA world, between certificate issues and parsing knots that `knife` is good with and `berks` isn't. Smoothing that over may need to wait for some other branch.

**Comment by amirahav** [comment]
Yes, uploading happens here: https://github.com/cloudamatic/mu/blob/min/bin/mu-upload-chef-artifacts#L140

Do you have happen to have a Berksfile that includes all the FEMA specific cookbooks/site-cookbooks?
Would it help if I wrote some code to dynamically create one if it doesn't exist?
Does it fail even after you first remove the local cache? It doesn't resolve dependencies for cookbooks that are not included in a Berksfile (Local cache exist only for cookbooks that aren't includes in a Berksfile).

**Comment by amirahav** [comment]
I'm open to reverting this commit and moving to a pull request...

**Comment by jstange** [comment]
Might have to, but I can't think straight right now, so let's figure that out later this week. For now I've bandaided my way around it by commenting out recipe funk and jamming certificates into Berksfiles.

**Comment by amirahav** [comment]
Cool, let me know. Not seeing issues with certificates here so I wonder what's going on

**Comment by jstange** [comment]
FEMA's Chef server uses a real SSL certificate instead of a self-signed one, which Berkshelf ironically doesn't seem to understand. No idea what cert stores it looks at, if any. My sketchy temporary workaround was to stick this in our Berksfiles:

`ENV['SSL_CERT_FILE'] = "/opt/mu/var/ssl/godaddy_chain.crt"`

**Comment by amirahav** [comment]
That's even weirder. I think I only tested it with a real cert.

**Comment by amirahav** [comment]
Yeah, so for self signed certs need to add _Mu_CA.pem_ to _/etc/ssl/certs/ca-bundle.crt_, and for real certs you may have to download a new ca-bundle from https://curl.haxx.se/docs/caextract.html

File: cookbooks/mu-tools/recipes/set_mu_hostname.rb [hunk 1]
```diff
@@ -16,22 +16,22 @@
...
-  if !node['ad']['computer_name']['nil? and !node']['ad']['computer_name'].empty?
+  if !node['ad']['computer_name'].nil? and !node['ad']['computer_name'].empty?
     $hostname = node['ad']['computer_name']
   end rescue NoMethodError
   $ipaddress = node['ipaddress']
   
   if !platform_family?("windows")
     my_deploy_id = nil
-    node.tags.map{ |key, value| my_deploy_id = value if key == 'MU-ID' } if node.respond_to?('tags')
+    node['tags'].map{ |key, value| my_deploy_id = value if key == 'MU-ID' } if node.respond_to?('tags')
```
jstange: Check out `get_tag`, which you'll get from the helper  library in `mu-tools`. We should drop all convoluted recipe code for getting `MU-*` tags in favor of that, imo.

amirahav:
For SSSD - yes, Samba is working with SSSD directly without winbind. I only used it with CentOS 7, however I would expect it to also work on CentOS 6.
Let me know if it works in FEMA, and I'll remove what needs to be removed from ```mu-tools::domain-node```.

cookbooks/mu-tools/recipes/set_mu_hostname.rb
amirahav: I think it should probably be called ```get_tag_value```, but : https://github.com/cloudamatic/mu/commit/724840eea55bcc1d22df375d4afe4df0af7e412d  (not tested)

jstange:
...ignore the message saying I approved the pull request. Playing with the new review functionality, some of it's not entirely intuitive.

I reverted `mu-upload-chef-artifacts` for now so I can get work done. @amirahav's changes are a good idea, we just need to iron them out, and that can wait for its own branch. We've got enough going on here.

modules/mu/clouds/aws/server.rb
```diff
@@ -1245,14 +1245,15 @@ def groom
...
           # Let us into any databases we depend on.
-          if @dependencies.has_key?("database")
-            @dependencies['database'].values.each { |db|
-              db.allowHost(@deploydata["private_ip_address"]+"/32")
-              if @deploydata["public_ip_address"]
-                db.allowHost(@deploydata["public_ip_address"]+"/32")
-              end
-            }
-          end
+          # This is probelmtic with autscaling - old ips are not removed, and access to the database can easily be given at the BoK level
```
jstange: Was planning to get snitty about this, but that's a good point. A better solution though: Allow members of the Autoscale group's automatically-generated security group instead of IPs. That gets around the cleanup problem. The moral equivalent would be what I just shoved into the BoK of this stack I'm working on:

```  ingress_rules:
  - port: 3306
    sgs:
    - poolrailsedu
```

...I'll teach the config parser that trick.

amirahav: My two cents - I really don't think it should be automatic, just like access control to other AWS resources - I think it should be more explicit and is very easy to do at the BoK level with SGs. This was the reason why I didn't switch this to using SGs...

jstange:
Dammit.. been having another weird mystery problem for the past week or so that I can't pin down. I'm seeing a race on resources that use a VPC created earlier in this deploy, in that they can't always look up a subnet created earlier in the deploy. I can't for the life of me see what's even in this branch that could cause this. It's erratic and hard to reproduce, but symptoms look like:

```
Oct 11 10:39:24 - /opt/mu/lib/modules/mu - Failed to locate subnet from {"subnet_name"=>"Public1"} in LoadBalancer lb
        {
  "thread": 45366760,
  "name": "loadbalancer_lb_create"
}
```

or

```
Oct 11 10:32:22 - deploy - #<NoMethodError: undefined method `private?' for nil:NilClass>
        {
  "thread": 36136360,
  "name": "database_db_create",
  "details": [
    "/opt/mu/lib/modules/mu/clouds/aws/database.rb:593:in `block in createSubnetGroup'",
```

amirahav:
So in not solving this but at least writing somewhat safer code - there should be a nil check here:
https://github.com/cloudamatic/mu/blob/min/modules/mu/clouds/aws/database.rb#L581
Because _getSubnet_ can return nil you may have an array of nil values.

Is it an issue with _loadSubnets_ not getting the correct (or any) _cloud_id_?

Do you send me the BoK? I can try looking at it as well...

jstange:
Yeah, I already have that nil check in my what-the-hell-is-this code. I suspect a general problem with subnet lookups. I'm seeing these things well after the subnets have been fully created and should be visible and available to the rest of the deploy. Notification may be bollocksed, or perhaps the lookup code in the VPC module.

jstange:
Yeah, there's some wild inconsistency across threads, and even within threads in different calls to `getSubnet`. Something smells. This doesn't seem related to work on this branch, but it's also only manifesting here, so I dunno what the heck. I'm chasing this down now, as it's a major blocker for ongoing work.

```
Oct 11 14:45:02 - clouds/aws/vpc - in VPC 'ICRAS-DEV-2016101114-RI-ICRASVPC' (vpc-5cac953b)'s getSubnet with cloud_id: , name: Public0
        {
  "thread": 46834440,
  "name": "loadbalancer_lb_create",
  "details": [
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public0' (subnet-70fd3e2b)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public1' (subnet-3602ce1b)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public2' (subnet-38cb7471)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Private2' (subnet-3bcb7472)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public3' (subnet-b836c984)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Private0' (subnet-77fd3e2c)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Private1' (subnet-3702ce1a)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Private3' (subnet-b936c985)"
  ]
}
Oct 11 14:45:02 - clouds/aws/vpc - in VPC 'ICRAS-DEV-2016101114-RI-ICRASVPC' (vpc-5cac953b)'s getSubnet with cloud_id: , name: Private0
        {
  "thread": 46836740,
  "name": "database_db_create",
  "details": [

  ]
}

Oct 11 14:45:03 - clouds/aws/vpc - in VPC 'ICRAS-DEV-2016101114-RI-ICRASVPC' (vpc-5cac953b)'s getSubnet with cloud_id: , name: Public1
        {
  "thread": 46834440,
  "name": "loadbalancer_lb_create",
  "details": [
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public0' (subnet-70fd3e2b)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public1' (subnet-3602ce1b)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public2' (subnet-38cb7471)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public1' (subnet-3602ce1b)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public2' (subnet-38cb7471)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public3' (subnet-b836c984)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public3' (subnet-b836c984)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Private0' (subnet-77fd3e2c)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Private1' (subnet-3702ce1a)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Private2' (subnet-3bcb7472)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Private0' (subnet-77fd3e2c)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Private3' (subnet-b936c985)"
  ]
}
Oct 11 14:45:03 - clouds/aws/vpc - in VPC 'ICRAS-DEV-2016101114-RI-ICRASVPC' (vpc-5cac953b)'s getSubnet with cloud_id: , name: Private1
        {
  "thread": 46836740,
  "name": "database_db_create",
  "details": [
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public0' (subnet-70fd3e2b)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public1' (subnet-3602ce1b)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public2' (subnet-38cb7471)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public3' (subnet-b836c984)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Private2' (subnet-3bcb7472)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Private3' (subnet-b936c985)"
  ]
}


Oct 11 14:45:03 - clouds/aws/vpc - in VPC 'ICRAS-DEV-2016101114-RI-ICRASVPC' (vpc-5cac953b)'s getSubnet with cloud_id: , name: Public2
        {
  "thread": 46834440,
  "name": "loadbalancer_lb_create",
  "details": [
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Public0' (subnet-70fd3e2b)",
    "Subnet 'ICRAS-DEV-2016101114-RI-ICRASVPC-Private3' (subnet-b936c985)"
  ]
}

Oct 11 14:50:57 - clouds/aws/database - Got nil subnet in list when attempting to build subnet group in Database
        {
  "thread": 38018460,
  "name": "database_db_create",
  "details": [
    "Subnet 'ICRAS-DEV-2016101114-WU-ICRASVPC-Private0' (subnet-b8fc3fe3)",
    null,
    null,
    "Subnet 'ICRAS-DEV-2016101114-WU-ICRASVPC-Private3' (subnet-9135caad)"
  ]
}

```

jstange:
Stupid `loadSubnets` needs to be thread-safe and just ain't. No idea how this has ever worked or why it only broke now. Fix incoming.

amirahav:
Does it only happen with inline VPCs ? I haven't seen this happen with any of the deployments into existing VPCs.

I can look at it later today, if you want a second pair of eyes and feel like committing your debug code.

amirahav:
kewl!

jstange:
Today I decide to burn a couple hours revisiting the weird problems exposed by exceptions in `MU::Clouds::AWS::DNSZone` and elsewhere, and... they've magically gone away. I can't see anything in commits we've made that's even relevant, so I don't know why it's fixed. I've uncommented those exceptions, since they exist for good reasons.

`(╯°□°)╯︵ ┻━┻`

Meanwhile, I also finished ripping most `CAPVolume` calls out of the `mu-tools` cookbook. I left the stuff dealing with device encryption in place. I **think** it's harmless as-is, the main point being that we don't want the node to have to make privileged API calls for itself. Significant regression testing merited.

I also changed `get_tag` to `mu_get_tag_value` per @amirahav feedback.

In general, I think it's time for @rpattcorner to start throwing this branch at his stuff. Remaining concerns, only one of which (the NAT recipe) is a big problem.

- Document the little support library that's growing in `cookbooks/mu-tools/libraries/helper.rb`, maybe use it to gloss over some repetitive Chef code elsewhere. (@jstange)
- Hide Scratchpad secrets behind some JavaScript so that peoples' dumb spam filters don't destroy secrets by prefetching them. (@zr2d2)
- Update or rewrite `mu-utility::nat` to use new firewall resources. This recipe is actually broken for unrelated reasons right now-  we need to explicitly add `ACCEPT` rules to the `FORWARD` chain for the network behind the NAT now, which wasn't necessary in the past. May as well modernize this recipe while we're at it.
- (OPTIONAL) Bang out a `StoragePool` implementation for the CloudFormation layer.

PR #51, opened 2016-07-20
Nested cloudformation

jstange:
Various enhancements to CloudFormation targets, and an armful of small bugfixes. Unlikely to impact other projects, but testing priority should be on #49 and #50 nonetheless.


jstange:
Oh, and a new feature I'd forgotten! Made a major enhancement to the config parser logic for VPCs that will, in the absence of instructions to the contrary, implicitly generate the standard set of subnets that we normally do anyway. In other words, this is now a complete VPC definition:

    - name: vpc
      route-tables:
      - name: internet
        routes:
        - destination_network: 0.0.0.0/0
          gateway: "#INTERNET"
      - name: private
        routes:
        - destination_network: 0.0.0.0/0
          gateway: "#NAT"

...which will give you a set of private and public subnets, one each for each local Availability Zone.

PR #52, opened 2016-08-12
fix existing db style

rpattcorner:
fix creationstyle=existing in combination with a dbid parameter to
identify the db.  Note, no guarantees on access if db is in another vpc


jstange:
Completed and merged. Dunno why Github didn't pick that up.

PR #61, opened 2016-12-26
more aggressive handling of Windows reboot irregularities

jstange:
Two changes here, which in theory should only impact Windows. We've been getting more weird failures bootstrapping Windows nodes, most of which seem to involve getting stuck and failing to reboot.

- Places in Chef where we trigger a reboot for Windows nodes are now using the Chef `reboot` resource. This is also set to trigger the `kill_ssh` helper or its equivalent, though that doesn't seem to help. I've removed certain things that try to trigger reboots themselves right before these calls, as that seems to confuse matters.

- When it inevitably gets stuck anyway, `MU::Cloud` still tries to force a Reboot from the API. Even that often fails now, so it has a second pass where it does a Stop, waits for it, and then Starts it again. If it's a node in an Autoscale group, it suspends scaling processes while that's happening.

@amirahav, if this muddles anything you're working on with 2016 or w/e, your feedback is encouraged.

### DIFF
--- a/cookbooks/mu-activedirectory/libraries/config.rb
+++ b/cookbooks/mu-activedirectory/libraries/config.rb
@@ -18,9 +18,18 @@ module Activedirectory
       Chef::Log.info("node_hostname: #{node.hostname.downcase}, computer_name: #{new_resource.computer_name.downcase}")
       if node.hostname.downcase != new_resource.computer_name.downcase
         cmd = powershell_out("Rename-Computer -NewName '#{new_resource.computer_name}' -Force -PassThru -Restart -DomainCredential #{creds}")
-        kill_ssh
         Chef::Application.fatal!("Failed to rename computer to #{new_resource.computer_name}") if cmd.exitstatus != 0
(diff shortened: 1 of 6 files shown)


amirahav:
@jstange let me look at that just to make sure it doesn't break anything here

amirahav:
@jstange let's hold off on this for a bit - the SSH session would generally get terminated correctly on restart, however it seems that isn't happening with these changes.

Also having issue with ssh not coming up correctly, which probably has nothing to do with the pull request - just Windows being Windows.

jstange:
sshd not terminating correctly is what this is a workaround for. Or do you mean the individual session getting dropped violently instead of gently?

amirahav:
I mean that the session doesn't always get terminated - the master will still get stuck on a session that has died, and it seems to be happening more often with this workaround.

jstange:
Aha. I've seen one instance of that, but it was before developing this, when I spent three days trying to hand-massage an ArcGIS node into being. Sounds like a tangentially-related problem, but not a causal relationship.

Maybe we add a timeout wrapper to certain sessions, specifically the ones where we know we're doing one finite operation that's expected to terminate (e.g. AD joins).

PR #70, opened 2017-10-05
Fixing stupid chef SSL issue

ryantiger658:
This issue should be fixed in the WinRM branch, but it is being problematic now. So I am pushing this fix.


jstange:
fixed in `demoUpdates` and `winrm_more_like_rm_windows`, dropping so we don't have merge conflicts with those guys.

PR #83, opened 2017-11-29
Issue 79

agill17:


agill17:
Turns out that the demo recipe (flask.rb) is depending on supervisor's default recipe which in chain depends on python cookbook. Unfortunately the supervisor cookbook has not been updated after python cookbook was deprecated. To work around this issue I will take down supervisors default recipe and other files/templates needed and update its resources to work with poise-python

agill17:
The demo flask recipe is updated here [78107eb](https://github.com/cloudamatic/mu/pull/84/commits/78107ebf67f8633e25fcc97aa72a7fe459f6c2b2)

jstange:
Merged.

Guys, in the future, try to use more descriptive titles than "issue_foo." I'm way too dumb to remember what each branch is about unless it says in the name.

PR #111, opened 2018-02-13
CFN Installer

agill17:


agill17:
@jstange I have added the govCloud base url issue in jira. And this is ready for review now. Thanks!

PR #114, opened 2018-03-09
The Goog

jstange:
Google Cloud layer
----------------------

We can now deploy `VPCs`, `FirewallRules`, `Servers`, `ServerPools`, and `LoadBalancers`[1] into Google Cloud Project. The following is a legal Basket of Kittens for deploying a VM in GCP:

```
---
  appname: foo
  servers:
  - name: simple
    cloud: Google
    platform: centos7
    size: g1-small
```

Mu Masters can reside in either of AWS or GCP, and can target either platform if appropriate credentials are available. Masters in GCP will transparently use the local machine's service account credentials to deploy resources, if no service account is explicitly provided. The default cloud for a given Mu Master will "whichever cloud this Master lives in." To create GCP resources from an AWS Master or vice versa, you will need to use the `cloud` parameter as in the above snippet.

API coverage for supported resource types is largely complete. We have attempted to keep the BoK language as transparently portable as possible, such that most well-written BoKs should Just Work on either platform[2]. Some options, however are simply GCP or AWS-specific, and the parser will identify them as such. `LoadBalancers` in particular have very different functionality in GCP.

To configure GCP support, use `mu-configure` to provide account credentials and a default target project (GCP accounts are subdivided into Projects, a useful privilege boundary).

```
12) Google Cloud Platform
  12a. Default Project - linear-theater-184819
  12b. Credentials Vault:Item - secrets:google
  12c. Default Region - us-east4
  12d. Log and Secret Bucket Name - stange-egtlabs-mu-dev
```

To create service account credentials, visit the GCP console at https://console.cloud.google.com/iam-admin/serviceaccounts/project and create a Service Account with `Project Owner` rights. The console will allow you to export credentials as a JSON file, which you will then import into a vault[3] so that Mu can use them (see the `Credentials Vault:Item` above). To import the JSON file to the vault as Mu expects to see it, use `knife vault create secrets google -J my-google-service-account.json`.

MU::Config refactoring
-------------------------

The last of our truly ugly modules, `MU::Config` has received a significant internal facelift. Much redundant code has been removed or condensed, and the ordering of validation has been altered. This should mostly be of interest for Mu developers.

Most significantly, schema and validation that is specific to a given cloud layer has been shifted out of `MU::Config` and into class methods in the appropriate cloud layer resource implementation. For example, LoadBalancers in GCP support an option called `named_ports`, which is one of the many possible targets for health checks and forwarding rules. Since this is cloud-specific, it is defined in `MU::Cloud::Google::LoadBalancer`.

```       
       def self.schema(config)
          toplevel_required = []
          schema = {
            "named_ports" => {
              "type" => "array",
              "items" => {
                "type" => "object",
                "required" => ["name", "port"],
                "additionalProperties" => false,
                "description" => "A named network port for a Google instance group, used for health checks and forwarding targets.",
                "properties" => {
                  "name" => {
                    "type" => "string"
                  },
                  "port" => {
                    "type" => "integer"
                  }
                }
              }
            }
          }
          [toplevel_required, schema]
        end
```

The `self.schema` method is a new required class method for all cloud resource implementations. It has a sibling, `self.validateConfig`, which is automatically called by the configuration parser to inspect for cloud-specific configuration errors, e.g. the following in `MU::Cloud::AWS::LoadBalancer.validateConfig`:

```
...
          if !lb["classic"]
            if lb["vpc"].nil?
              MU.log "LoadBalancer #{lb['name']} has no VPC configured. Either set 'classic' to true or configure a VPC.", MU::ERR
              ok = false
            end
...
```

These are significant structural changes with a lot of breakage potential, and will require extensive testing before this branch is considered release-ready.

There is much of our standard generic config schema that is, probably, AWS-specific, and should be migrated into AWS modules over time. This is not of immediate urgency, however.

BoK params exposed to Chef
--------------------------------

If your Basket of Kittens has any `parameters` in it, these and their values will be exposed to Chef recipes on your nodes in `node['mu_params']`. Handy.

Developer FAQ for cloud resource implementations
---------------------------------------------------------

There is now a [README](https://github.com/cloudamatic/mu/tree/the_goog/modules/mu/clouds) describing the requirements for implementing a new cloud resource.

us_only BoK flag for many resources
----------------------------------------

Restrict regions for target resources to US-only, valuable for government tasks. This is primarily of use in Google Cloud, where many resources can span regions globally (e.g. `VPCs`). The flag is supported in the AWS layer as well, in case this is of use to cloud layer developers.

AWS GovCloud support
--------------------------

Idiosyncratic behaviors related to AWS GovCloud regions should now be smoothed over. Behavior should be largely be similar to the commercial regions, though many services are not supported by Amazon in GovCloud (notably Route53, aka `DNSZones`).

mu-upload-chef-artifacts no longer deletes things by default
-------------------------------------------------------------------

You don't have to run `mu-upload-chef-artifacts` with `-n` anymore to keep it from wasting five mintus purging the Chef server before (re)uploading everything. Default behavior now skips this purging step. If you would like it to purge things, use the `-p` option.

CAPVolume Chef library removed
--------------------------------------

Deprecated for some time, this has been fully removed from our Chef ecosystem. The `mu_tools_disk` resource should be used, if it is necessary to create volumes via Chef recipes.

Mu Masters no longer create their own log volumes via Chef, but rather through the `mu-aws-setup` or `mu-gcp-setup` utility, whichever is applicable. They in turn call the `MU::Master.disk` utility method, which retains the old functionality for encrypted volumes with `luks`. Right now this is only used for `/Mu_Logs`.

AWS AutoScale Scheduled Actions
--------------------------------------

AutoScale can now adjust the `min_size`, `max_size`, and/or `desired_capacity` on a one-off basis or a regular, cron-like schedule. This is now supported in BoK language: http://35.153.108.101/docs/MU/Config/BasketofKittens/server_pools/schedule.html

```
  schedule:
  - action_name: scale-down-over-night
    recurrence: "0 23 * * *"
    min_size: 1
    max_size: 1
  - action_name: scale-up-during-the-day
    recurrence: "0 10 * * *"
    min_size: 3
    max_size: 3
```

This can be added to existing deployments with `-u` and the experimental `-l` flag in combination. Example:

```
$ mu-deploy mystack.yaml -u MYSTACK-DEV-2018031216-KU -l
<...stuff...>
Mar 16 16:35:58 - clouds/aws/server_pool - Adding scheduled action to AutoScale group MYSTACK-DEV-2018031216-KU-SVR
{:auto_scaling_group_name=>"MYSTACK-DEV-2018031216-KU-SVR",
 :scheduled_action_name=>"scale-down-over-night",
 :max_size=>1,
 :min_size=>1,
 :recurrence=>"0 20 * * *"}
Mar 16 16:35:58 - clouds/aws/server_pool - Adding scheduled action to AutoScale group MYSTACK-DEV-2018031216-KU-SVR
{:auto_scaling_group_name=>"MYSTACK-DEV-2018031216-KU-SVR",
 :scheduled_action_name=>"scale-up-during-the-day",
 :max_size=>2,
 :min_size=>2,
 :recurrence=>"0 7 * * *"}
```

mu-node-manage -m chefupgrade
---------------------------------------

Nodes in long-standing deployments often lag far behind on Chef Client versions, to a point where some community cookbooks cease working. This new `mu-node-manage` subcommand will jump in and replace the installed client with Mu's current default client version.

Chef Server and Client updates
----------------------------------

Chef Client default: `12.21.31-1`
Chef Server default: `12.17.15-1`

As always, Chef Server upgrades are known to be dodgy (Opscode failsauce, usually around rabbitmq). Approach with caution. Recommended procedure:

1. If the machine contains production deployments or other important data, image it beforehand for the sake of recoverability
2. `mu-self-update -b the_goog`
3. The Chef Server upgrade will probably fail outright or hang indeterminately. If it does not, you're done, and can skip the rest of these steps.
4. `chef-server-ctl stop && rpm -e chef-server-core && reboot`
5. After a reboot, run `service iptables stop`, because sometimes a background Chef run will get confused and block a bunch of ports on you.
6. `chef-apply /opt/mu/lib/cookbooks/mu-master/recipes/init.rb` to reinstall the Chef Server package and configure it appropriately. Your pre-existing Chef data should remain and not require restoration.
7. Run `mu-self-update -b the_goog` again to complete the upgrade

Misc Bugfixes
---------------

- Myriad RHEL7-family 389ds fixes. Workarounds for weird behavior from the package's installer and for missing SELinux permissions.
- Further low-key syntax updates to cookbook ecosystem in preparation for Chef 13/14 compatibility
- Display banners correctly and set hostname correctly on RHEL7-family nodes and masters.
- Mysterious mis-setting of the `nodename` field in server deployment metadata may be fixed; existing deploys with such mangled metadata should self-correct on next groom

[1] GCP `Databases`, e.g. Cloud SQL (the equivalent of AWS RDS) are not supported yet, due to API issues. This is almost certainly the next implementation target.

[2] As of this writing the `size` parameter doesn't automatically figure out what you mean when, say, you ask for a `t2.micro` instance (an AWS machine type) in GCP. We think we'll come up with some magic for this before release, in keeping with our philosophy of maximum universality across cloud providers.

[3] This mechanism is better than the plain-text storage we use for AWS credentials currently, and we expect to backport the vault functionality to the AWS layer at some point in the near future.


jstange:
TODO before this goes live: Slip in Scheduled Events for AWS Autoscaling. It'll be an easy, large win.

https://jira.egt-labs.com/browse/CLOUDMATIC-96

ryantiger658:
I give a green light to merging this into master. 

giphy 2

PR #115, opened 2018-03-12
Fixup inspec to use deploy ids instead of BOK 

rpattcorner:
So ... can examine deploys from anywhere now, not just in demos
And ... can distinguish between multiple deploys that share some name in common

Not quite sure what the changes in init.rb and installer that list the branch name are ... need to back out?


agill17:
Tested and works! Merging.

PR #116, opened 2018-03-13
CFN Installer needs EIP and General Purpose SSD for root EBS

agill17:


jstange:
Seems reasonable. I'll merge manually to deal with the branch name.

PR #136, opened 2018-06-19
Installer bugfix

tzthib:
We fixed a few bugs in the installer, updated the demo platform to point to GitHub, and updated the grep filter for CentOS7.


ryantiger658:
CodeClimate errors are trying to run rubocop on chef cookbooks and as a result are not happy. We should look into those settings to see if we can tell it not to run on the cookbooks.

PR #139, opened 2018-07-23
Nuisances

jstange:
- `StoragePool` security groups should correctly associate with mountpoints
- Handle EC2 Classic Mu Master security group correctly
- Fix validator on `Notification`
- Fix validator on `Alarm`
- Somewhat more sensible defaulting on `Database` resources

### DIFF
--- a/bin/mu-configure
+++ b/bin/mu-configure
@@ -87,7 +87,6 @@ $CONFIGURABLES = {
     "changes" => ["chefartifacts", "chefrun"],
     "array" => true,
     "default" => ['https://github.com/cloudamatic/mu_demo_platform']
-    "default" => []
   },
   "master_runlist_extras" => {
     "title" => "Mu Master Runlist Extras",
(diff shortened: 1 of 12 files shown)


ryantiger658:
@jstange is there anything we can do to fix the 4 new Codeclimate issues?

jstange:
Fixed the one that was an actual problem. Its other complaints are stupid.

ryantiger658:
Sounds good. Just wanted to make sure we are looking at them.

jstange:
Yeah, we should be, as much as I want to grump about it. We still need to sit down and tune Rubocop to some best practices defined by grownups. Far as I can tell its defaults are geared toward Rails bootcamp students.

ryantiger658:
I have a branch fixing (most) of the Chef Foodcritic problems. I just need to get time to test and merge. 

I am hoping someone a little more versed in ruby will take point on the Rubocop issues.

ryantiger658:
Closes CLOUDMATIC-160

PR #141, opened 2018-08-02
better error handling on nonsensical RDS behavior

jstange:
Fix some database-specific parser bugs. Also respond better to one very peculiar edge case:

If you try to build an RDS instance from a clone of another (`creation_style: existing`), AWS will quietly put the resulting database into the same VPC as the source database, even if you requested a different one. This is deeply confusing, so now `MU::Cloud::AWS::Database` will throw an error, explain what's going on, and advise you to use `new_snapshot` instead.


ryantiger658:
@jstange please comment or fix the codeclimate stuffs. 

Thanks!

jstange:
CodeClimate can suck my rocket in this particular instance

ryantiger658:
Ran this on random weird BoK's and am not seeing any errors. I do want to use this to create simple and complex tests for the database resources.

PR #142, opened 2018-08-10
Fix Foodcritic Warnings

ryantiger658:
I went through all of our cookbooks and fixed many of our foodcritic issues. 

*THIS NEEDS SIGNIFICANT TESTING* This work effects all of our cookbooks. 

Requesting a merge into development so that we can spend some time testing this thoroughly.


ryantiger658:
We have merged this into `chef_14_support` this PR is no longer relevant.

PR #160, opened 2019-04-24
fix some upgrade-related irritations

jstange:


jstange:
@ryantiger658 This is ready for attention now. We think the dust is settled.

PR #167, opened 2019-09-05
Revert "workaround for deprecated gem option"

zr2d2:
Reverts cloudamatic/mu#166


zr2d2:
the original merge incremented it, so `git revert` undid it

zr2d2:
I already did it locally

PR #170, opened 2019-09-30
add in mu-master cookbook dependencies

zr2d2:
add in chef-sugar dependency to work around missing dependency from `consul` cookbook


zr2d2:
later versions of `consul` don't require chef-sugar, but the things that rely on `consul` won't let me upgrade it

PR #175, opened 2019-11-19
Cicd container deploy

zr2d2:
add ability to build docker containers and push them to private and public registries


.gitlab-ci.yml
```diff
@@ -5,14 +5,17 @@ stages:
...
+  - Docker Build
   - Smoke Test
   - Merge/Tag
-  - Dockerbuild
   - Deploy
   
 Rubocop:
   stage: Lint Test
   image: chef/chefdk:latest
+  variables: {
```
ryantiger658: Can you switch this syntax to the following?

```
variables:
     variable_1: value
     variable_2: value
````

It will be more consistent with our style

File: Dockerfile [hunk 2]
```diff
@@ -0,0 +1,20 @@
+FROM centos:7
```
ryantiger658: Did we switch the base image to centos? For some reason I thought we were building off of ruby2.5 base

modules/mu/clouds/aws/server.rb
```diff
@@ -120,6 +120,7 @@ def initialize(**args)
             @config['mu_name'] = @mu_name
 
           end
+          @groomer = MU::Groomer.new(self) unless MU.inGem?
```
ryantiger658: @jstange do we have ansible bootstraping in 3.0.0? Or are we still dependent on chef for bootstrap?

modules/mu/clouds/aws/server.rb
```diff
@@ -938,11 +939,17 @@ def postBoot(instance_id = nil)
 
           # See if this node already exists in our config management. If it does,
           # we're done.
-          if @groomer.haveBootstrapped?
+          if MU.inGem?
+            MU.log "Deploying from a gem, not grooming"
```
ryantiger658: Good communication with the users. :)

Dockerfile
zr2d2: The ruby 2.5 container is Debian based, which is not yet supported, so I changed it so the build environment was closer to a supported production environment

PR #176, opened 2019-11-20
basic test kitchen

zr2d2:
add test kitchen to pipeline. Spins up Amazon Linux and Centos 7 machines and tests the installer against it


modules/Gemfile.lock
```diff
@@ -43,7 +43,7 @@ GEM
       public_suffix (>= 2.0.2, < 4.0)
     ast (2.4.0)
     aws-eventstream (1.0.3)
-    aws-sdk-core (2.11.395)
```
ryantiger658: If these are incorrect I blame @zr2d2.

